PR #84, opened 2015-03-06
GIS: Possible fix for barcode sometimes not showing up on initial boot

sgnn7:
While we can't pinpoint the reason for the barcode disappearing, the
best guess is that sometiumes the cache dir is being created by some
other method that might encounter a race condition with GIS so this fix
makes sure that we have that directory before trying to write the
barcode image to it.

[endlessm/eos-shell#4708]

### DIFF
--- a/gnome-initial-setup/pages/language/gis-language-page.c
+++ b/gnome-initial-setup/pages/language/gis-language-page.c
@@ -209,9 +209,15 @@ static gchar *
 create_serial_barcode (const gchar *serial)
 {
   gchar *savefile;
+  const gchar *cache_dir;
   struct zint_symbol *barcode;
 
-  savefile = g_build_filename (g_get_user_cache_dir (), "product_serial.png", NULL);
+  cache_dir = g_get_user_cache_dir ();
+
+  /* Create the directory if it's missing */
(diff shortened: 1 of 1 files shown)


gnome-initial-setup/pages/language/gis-language-page.c
```diff
@@ -209,9 +209,17 @@ static gchar *
...
+  const gchar *cache_dir;
   struct zint_symbol *barcode;
 
-  savefile = g_build_filename (g_get_user_cache_dir (), "product_serial.png", NULL);
+  cache_dir = g_get_user_cache_dir ();
+
+  /* Create the directory if it's missing */
+  if (!g_file_test (cache_dir, G_FILE_TEST_IS_DIR)) {
+    g_mkdir (cache_dir, 0755);
+  }
```
cosimoc: There is no need to call `g_file_test()` here beforehand.
You can just call `g_mkdir_with_parents()` and ignore the return value.

sgnn7: @cosimoc Oh, cool. I'll make the change.

sgnn7:
@cosimoc Changed to use `g_mkdir_with_parents`

PR #87, opened 2015-03-19
Add a music player

magcius:
Add the ability to play some chill tunes in a loop while going through
the FBE. There's a minimal rampout fade, but it's not hooked up yet.

We aren't going to install the actual music asset yet, pending approval
our creative team. Thankfully, it's super simple to drop in once we do
approve it.

[endlessm/eos-shell#4765]


gnome-initial-setup/gis-driver.c
```diff
@@ -85,6 +86,8 @@ struct _GisDriverPrivate {
   gchar *lang_override;
   gchar *default_timezone;
 
+  GisMusicPlayer *music_player;
```
cosimoc: Nothing ever frees this

gnome-initial-setup/gis-driver.c
```diff
@@ -312,6 +316,7 @@ gis_driver_startup (GApplication *app)
 
   priv->main_window = gis_window_new (driver);
   priv->assistant = gis_window_get_assistant (GIS_WINDOW (priv->main_window));
+  priv->music_player = gis_music_player_new ("file://" PKGDATADIR "/fbe.ogg");
```
cosimoc: Do we have `fbe.ogg` as part of this PR? It's fine if we don't, as long as the music player gracefully fails.
Also I wonder if we should not put the music into a GResource like we do for everything else

magcius: It is not, pending approval from Dalio. I can put in the temporary asset for now.

cosimoc: I think it's fine to leave the asset out for now, as long as it doesn't crash if the asset is not there

gnome-initial-setup/gis-music-player.c
```diff
@@ -0,0 +1,216 @@
...
+G_DEFINE_TYPE_WITH_PRIVATE (GisMusicPlayer, gis_music_player, G_TYPE_OBJECT);
+
+static void
+gis_music_player_dispose (GObject *object)
+{
+  GisMusicPlayer *player = GIS_MUSIC_PLAYER (object);
+  GisMusicPlayerPrivate *priv = gis_music_player_get_instance_private (player);
+
+  g_clear_object (&priv->bus);
+  g_clear_object (&priv->playbin);
```
cosimoc: Should also remove the `fade_timeout_id`
Also should chain up to parent

gnome-initial-setup/gis-music-player.c
```diff
@@ -0,0 +1,216 @@
...
+      message->src == GST_OBJECT (priv->playbin))
+    {
+      gst_element_seek_simple (priv->playbin, GST_FORMAT_TIME,
+                               GST_SEEK_FLAG_SEGMENT,
+                               priv->loop_point);
+    }
+}
+
+GisMusicPlayer *
+gis_music_player_new (const char *uri)
```
cosimoc: The uri should be a property of the object...
So this code should probably be moved either in the property setter or in one of the initialization functions

gnome-initial-setup/gis-music-player.c
```diff
@@ -0,0 +1,216 @@
...
+
+  return player;
+}
+
+void
+gis_music_player_play (GisMusicPlayer *player)
+{
+  GisMusicPlayerPrivate *priv = gis_music_player_get_instance_private (player);
+  priv->playing = TRUE;
+  gst_element_set_state (priv->playbin, GST_STATE_PAUSED);
```
cosimoc: Shouldn't this be `GST_STATE_PLAYING`?

cosimoc: Also should probably cancel the fadeout if this is called again

gnome-initial-setup/gis-music-player.c
```diff
@@ -0,0 +1,216 @@
...
+    {
+      g_object_set (priv->playbin, "volume", 0.0, NULL);
+      priv->fade_timeout_id = 0;
+      gst_element_set_state (priv->playbin, GST_STATE_PAUSED);
+      return FALSE;
+    }
+}
+
+void
+gis_music_player_rampout (GisMusicPlayer *player)
```
cosimoc: What happens when this is called and the rampout is already going on?

gnome-initial-setup/gis-music-player.c
magcius: No, it should be `PAUSED`. See the `on_state_changed` which sets it to PLAYING after the preroll is done. We need to hold off in order to do a proper segment seek.

gnome-initial-setup/gis-music-player.c
magcius: yeah but properties are stupidly hard to write and i hate them. give me gproperty

gnome-initial-setup/gis-music-player.c
cosimoc: Actually scrap my comment about `PLAYING. I see what you do in the callback

wjt:
I don't think we're doing this (though we are playing a startup sound on some images: #268 ).

PR #94, opened 2015-05-29
Factory test mode

dbnicholson:
Wire up a script to mount scratch overlays for a test mode.

[endlessm/eos-shell#5036]


dbnicholson:
CR ready again. I think I addressed the comments from before, although they've been lost into the ether since I amended the commits.

PR #96, opened 2015-06-02
account: Delete root password after user created

dbnicholson:
Endless ships with a stock password for root in order to allow early
debugging of systems. Delete this password after the new user has been
created.

As a precaution, first check if the system is booted into an ostree by
looking for the ostree= option in /proc/cmdline. This is how ostree
itself determines it.

[endlessm/eos-shell#5123]


wjt:
Endless no longer ships with a default root password.

PR #100, opened 2015-06-18
build: bump webkit dependency version

cosimoc:
Depend on webkit2gtk-4.0 instead of webkit2gtk-3.0.

[endlessm/eos-shell#5202]


dbnicholson:
That's all it takes, huh?

PR #110, opened 2015-11-25
Don't run first boot experience for new user accounts

mcatanzaro:
Run FBE only for the very first user account on the system, the first
time the system boots, and never for new user accounts.

[endlessm/eos-shell#3727]

### DIFF
--- a/data/Makefile.am
+++ b/data/Makefile.am
@@ -56,9 +56,9 @@ CLEANFILES =						\
 	$(NULL)
 
 autostartdir = $(sysconfdir)/xdg/autostart
+# Endless does not install gnome-initial-setup-first-login.desktop
 autostart_DATA =					\
 	gnome-initial-setup-copy-worker.desktop		\
-	gnome-initial-setup-first-login.desktop		\
 	$(NULL)
 
 tmpfilesdir = $(libdir)/tmpfiles.d


mcatanzaro:
Hm, do you want me to rip out all traces of this functionality? I made the smallest-possible change thinking we might want to upgrade gnome-initial-setup in the future, so having a small diff would be best.

Maybe I should add comments to the Makefile to explain why we have removed the desktop file.

data/Makefile.am
```diff
@@ -51,14 +50,12 @@ CLEANFILES =						\
...
 	gnome-initial-setup-copy-worker.desktop.in	\
 	gnome-initial-setup-copy-worker.desktop		\
-	gnome-initial-setup-first-login.desktop		\
 	com.endlessm.TestMode.policy			\
 	$(NULL)
 
 autostartdir = $(sysconfdir)/xdg/autostart
 autostart_DATA =					\
 	gnome-initial-setup-copy-worker.desktop		\
-	gnome-initial-setup-first-login.desktop		\
```
cosimoc: @mcatanzaro I think just avoiding the installation of the desktop file is a good approach, but I think the only change that's needed in that case is the removal of this line.
I actually would prefer to comment it out with a note like you propose.

PR #114, opened 2015-12-11
Issue/5924

GeorgesStavracas:
This PR fixes the following issues:
- Makes the window adapt to small screen sizes.
- Adapt the layout to fit smaller sizes.

[endlessm/eos-shell#5924]

### DIFF
--- a/gnome-initial-setup/gis-driver.c
+++ b/gnome-initial-setup/gis-driver.c
@@ -71,7 +71,7 @@ enum {
 static GParamSpec *obj_props[PROP_LAST];
 
 struct _GisDriverPrivate {
-  GtkWidget *main_window;
+  GtkWindow *main_window;
   GisAssistant *assistant;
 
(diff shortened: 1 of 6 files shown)


gnome-initial-setup/gis-driver.c
```diff
@@ -98,6 +98,43 @@ gis_driver_finalize (GObject *object)
   G_OBJECT_CLASS (gis_driver_parent_class)->finalize (object);
 }
 
+static void
+prepare_main_window (GisDriver *driver)
+{
+  GisDriverPrivate *priv = gis_driver_get_instance_private (driver);
+  GdkGeometry size_hints;
+
+  if (gis_driver_is_small_screen (driver))
```
cosimoc: The function does not take the driver anymore as an argument, so you should remove this from here (I'm surprised it builds without warnings!)

PR #117, opened 2016-01-07
Issue/5924

GeorgesStavracas:
This pull requests presents the following changes:
- Fix the margin on Location page
- Adjust labels thoughout pages so they don't resize the window to enormous widths
- Fix the scrollbar showing up on bottom when FBE starts

[endlessm/eos-shell#5924]

### DIFF
--- a/gnome-initial-setup/gis-driver.c
+++ b/gnome-initial-setup/gis-driver.c
@@ -140,7 +140,7 @@ prepare_main_window (GisDriver *driver)
       gtk_window_set_geometry_hints (priv->main_window,
                                      NULL,
                                      &size_hints,
-                                     GDK_HINT_MIN_SIZE | GDK_HINT_WIN_GRAVITY);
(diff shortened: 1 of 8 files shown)


gnome-initial-setup/gis-driver.c
```diff
@@ -87,6 +87,14 @@ typedef struct _GisDriverPrivate GisDriverPrivate;
 G_DEFINE_TYPE_WITH_PRIVATE(GisDriver, gis_driver, GTK_TYPE_APPLICATION)
 
 static void
+scrolled_window_size_allocate (GtkScrolledWindow *sw,
+                               GdkRectangle      *alloc,
+                               GisDriver         *driver)
+{
+  gtk_adjustment_set_value (gtk_scrolled_window_get_vadjustment (sw), 0);
```
cosimoc: This does not look entirely right:
- it should probably use `gtk_adjustment_get_lower()` instead of `0`
- it does not seem right to do this every size-allocate

Can you not do this only when the page is switched?

PR #119, opened 2016-02-11
input-chooser: don't access priv pointer before checking no error

cosimoc:
If the GCancellable is cancelled by the finalize() function, the
callback will still be called potentially with an instance of the
CcInputChooser that's not valid anymore.
To avoid crashing in that case, only access the priv pointer of
CcInputChooser when we have no error.

[endlessm/eos-shell#6362]

### DIFF
--- a/gnome-initial-setup/pages/keyboard/cc-input-chooser.c
+++ b/gnome-initial-setup/pages/keyboard/cc-input-chooser.c
@@ -621,18 +621,21 @@ fetch_ibus_engines_result (GObject       *object,
                            GAsyncResult  *result,
                            CcInputChooser *chooser)
 {
-        CcInputChooserPrivate *priv = cc_input_chooser_get_instance_private (chooser);
+        CcInputChooserPrivate *priv;
         GList *list, *l;
         GError *error;
 
         error = NULL;
-        list = ibus_bus_list_engines_async_finish (priv->ibus, result, &error);
-        g_clear_object (&priv->ibus_cancellable);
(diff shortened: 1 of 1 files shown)


gnome-initial-setup/pages/keyboard/cc-input-chooser.c
```diff
@@ -621,19 +621,22 @@ fetch_ibus_engines_result (GObject       *object,
...
         if (!list && error) {
-                g_warning ("Couldn't finish IBus request: %s", error->message);
+                if (!g_error_matches (error, G_IO_ERROR, G_IO_ERROR_CANCELLED))
+                        g_warning ("Couldn't finish IBus request: %s", error->message);
                 g_error_free (error);
                 return;
         }
 
+        priv = cc_input_chooser_get_instance_private (chooser);
+        g_clear_object (&priv->ibus_cancellable);
```
mcatanzaro: The last ref on the cancellable is held by the GTask, and it will be freed once you return, correct?

cosimoc: There is no GTask here - the cancellable is created before calling in the async method - see https://github.com/endlessm/gnome-initial-setup/blob/shell/6362/gnome-initial-setup/pages/keyboard/cc-input-chooser.c#L666

mcatanzaro: The GTask is the GAsyncResult ;)

Anyway, I'm pretty sure this is right.

PR #123, opened 2016-03-22
Initialize locales in existing user mode

mcatanzaro:
Since we don't run GTK+ anymore, we have to do this manually, else the
tutorial will always start in English.

https://phabricator.endlessm.com/T3727

### DIFF
--- a/gnome-initial-setup/gnome-initial-setup.c
+++ b/gnome-initial-setup/gnome-initial-setup.c
@@ -315,6 +315,7 @@ main (int argc, char *argv[])
   /* Upstream has "existing user" mode for new user accounts. In Endless, we
      skip straight to the welcome tutorial instead. */
   if (get_mode () == GIS_DRIVER_MODE_EXISTING_USER) {
+    setlocale (LC_ALL, "");
     launch_tutorial ();
     gis_add_setup_done_file ();
     return EXIT_SUCCESS;


gnome-initial-setup/gnome-initial-setup.c
```diff
@@ -315,6 +315,7 @@ main (int argc, char *argv[])
   /* Upstream has "existing user" mode for new user accounts. In Endless, we
      skip straight to the welcome tutorial instead. */
   if (get_mode () == GIS_DRIVER_MODE_EXISTING_USER) {
+    setlocale (LC_ALL, "");
```
cosimoc: In retrospect I would have preferred it if we modified the `GisDriver` to run with this new behavior when in existing user mode, instead of doing it behind its back, but I guess it's too late for that now. We can revisit that the next time we rebase on a new g-i-s.

However, it seems more robust to at least move this whole block after the GTK initialization.

mcatanzaro: If you want, but I'm not sure that's best. For an upstream patch, we'd of course want a much more thorough change, but for Endless, don't you think it's probably best to quit early in main, before running GisDriver, since we don't need any of its functionality and to minimize our delta?

Also, since we don't use GTK+ at all, I don't see why we should move this code beneath GTK+ initialization. It seems like overkill to initialize GTK+ just to get it to call setlocale() for us. :)

cosimoc: I won't insist on this point, given that we're really close to the end game for 2.6.0.
But my concern is that we later do something else that requires us to initialize GTK and we have to do that anyway (or we're bitten by it because we don't realize we should have initialized it...)

PR #133, opened 2016-10-10
T12547

GeorgesStavracas:
This PR implements the latest comments on issue T12547

https://phabricator.endlessm.com/T12547

### DIFF
--- a/gnome-initial-setup/pages/live-chooser/gis-live-chooser-page.ui
+++ b/gnome-initial-setup/pages/live-chooser/gis-live-chooser-page.ui
@@ -52,9 +52,6 @@
                     <attributes>
                       <attribute name="scale" value="1.4"/>
                     </attributes>
-                    <style>
-                      <class name="dim-label"/>
-                    </style>
                   </object>
                   <packing>
                     <property name="expand">False</property>
@@ -128,9 +125,6 @@
                     <attributes>
                       <attribute name="scale" value="1.4"/>
                     </attributes>
-                    <style>
-                      <class name="dim-label"/>
-                    </style>
                   </object>
                   <packing>
                     <property name="expand">False</property>
--- a/gnome-initial-setup/pages/live-chooser/gis-live-chooser-page.ui
+++ b/gnome-initial-setup/pages/live-chooser/gis-live-chooser-page.ui
@@ -189,7 +189,7 @@
   <object class="GtkSizeGroup">
     <property name="mode">vertical</property>
     <widgets>
-      <widget name="install_icon"/>
+      <widget name="reformat_icon"/>
       <widget name="try_icon"/>
     </widgets>
   </object>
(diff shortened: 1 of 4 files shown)


gnome-initial-setup/pages/location/gis-location-page.c
```diff
@@ -812,6 +812,9 @@ gis_location_page_init (GisLocationPage *page)
 void
 gis_prepare_location_page (GisDriver *driver)
 {
+  if (gis_driver_is_live_session (driver))
+    return;
```
wjt: The commit message talks about not wanting to update the hardware clock, but does this actually achieve that goal?

I've read the [timedated](https://www.freedesktop.org/wiki/Software/systemd/timedated/) spec and some of its source, and it seems that it will only update the clock when explicitly asked to, or when one calls `SetLocalRTC` with the `fix_system` parameter set to `false`. The spec seems to suggest calling `SetLocalRTC(local_rtc=true, fix_system=true, interactive=false)` in our situation:

> Use fix_system=true in installers and livecds where the RTC is probably more reliable than the system time.

though maybe we can leave well alone and hope it'll be okay. Did you look into this?

gnome-initial-setup/pages/live-chooser/gis-live-chooser-page.ui
```diff
@@ -189,7 +183,7 @@
   <object class="GtkSizeGroup">
     <property name="mode">vertical</property>
     <widgets>
-      <widget name="install_icon"/>
+      <widget name="reformat_icon"/>
```
wjt: Good catch.

gnome-initial-setup/pages/location/gis-location-page.c
GeorgesStavracas: Do we want to hide the page, or display it and make it call SetLocalRTC()?

wjt: We definitely want to hide it. If feasible, I would like to *also* avoid changing the hardware clock. But, if booting a current Endless OS live USB on a Windows machine (with a non-UTC timezone) doesn't mess with the clock, just hiding the page is probably fine. (I don't have one to test on ☹)

gnome-initial-setup/pages/location/gis-location-page.c
```diff
@@ -687,6 +687,25 @@ gis_location_page_constructed (GObject *object)
 
   update_timezone (page);
 
+  /*
+   * When running on a Live session, it's important to try to make the best
+   * usage of the system clock. Timedate1 has a handy method for that.
```
wjt: rather than "best usage" and "handy method", can we describe what this call achieves?

gnome-initial-setup/pages/location/gis-location-page.c
```diff
@@ -687,6 +687,25 @@ gis_location_page_constructed (GObject *object)
...
+                                       TRUE,
+                                       TRUE,
+                                       FALSE,
+                                       NULL,
+                                       &error);
+
+    if (error) {
+      g_warning ("Error calling SetLocalRTC(): %s", error->message);
+      g_clear_error (&error);
+      exit (1);
```
wjt: Let's not terminate the whole process if calling this method fails… It is not the end of the world if this doesn't work, and if this path is reached for some unexpected reason then our users will have totally unusable live USBs and will send us angry emails.

PR #136, opened 2016-10-27
eula: disable metrics on live systems

wjt:
https://phabricator.endlessm.com/T13890


wjt:
See https://github.com/endlessm/eos-metrics-instrumentation/pull/62 for another approach to metrics on live systems.

PR #142, opened 2016-11-16
network: skip page when ethernet cable is available

GeorgesStavracas:
When showing the Network page on the initial setup,
the list of Wi-Fi connections is displayed even when
there's an ethernet cable available.

Fix that by skipping the page when any ethernet device
if detected.

https://phabricator.endlessm.com/T13749

### DIFF
--- a/gnome-initial-setup/pages/network/gis-network-page.c
+++ b/gnome-initial-setup/pages/network/gis-network-page.c
@@ -599,6 +599,7 @@ gis_network_page_constructed (GObject *object)
   DBusGConnection *bus;
   GError *error;
   gboolean visible = TRUE;
+  gboolean has_ethernet_connection = FALSE;
   GtkWidget *box;
 
   G_OBJECT_CLASS (gis_network_page_parent_class)->constructed (object);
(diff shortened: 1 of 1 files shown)


gnome-initial-setup/pages/network/gis-network-page.c
```diff
@@ -617,14 +618,28 @@ gis_network_page_constructed (GObject *object)
       if (!nm_device_get_managed (device))
         continue;
 
-      if (nm_device_get_device_type (device) == NM_DEVICE_TYPE_WIFI) {
+      switch (nm_device_get_device_type (device)) {
+      case NM_DEVICE_TYPE_ETHERNET:
+        has_ethernet_connection = TRUE;
+        priv->nm_device = g_object_ref (device);
```
mcatanzaro: This only works if you have just one NMDevice, but this is a loop that handles multiple NMDevices. So in the common case of ethernet plus wi-fi device, you leak whichever one gets processed first. At minimum you need a g_clear_object in both cases, but probably you don't really need to set priv->nm_device if it's Ethernet at all, right...?

PR #143, opened 2016-11-21
T13749

GeorgesStavracas:
https://phabricator.endlessm.com/T13749

### DIFF
--- a/gnome-initial-setup/pages/network/gis-network-page.c
+++ b/gnome-initial-setup/pages/network/gis-network-page.c
@@ -599,7 +599,6 @@ gis_network_page_constructed (GObject *object)
   DBusGConnection *bus;
   GError *error;
   gboolean visible = TRUE;
-  gboolean has_ethernet_connection = FALSE;
   GtkWidget *box;
 
   G_OBJECT_CLASS (gis_network_page_parent_class)->constructed (object);
@@ -618,24 +617,12 @@ gis_network_page_constructed (GObject *object)
       if (!nm_device_get_managed (device))
         continue;
 
-      switch (nm_device_get_device_type (device)) {
-      case NM_DEVICE_TYPE_ETHERNET:
-        has_ethernet_connection = TRUE;
-        break;
-
-      case NM_DEVICE_TYPE_WIFI:
+      if (nm_device_get_device_type (device) == NM_DEVICE_TYPE_WIFI) {
         /* FIXME deal with multiple, dynamic devices */
-        g_set_object (&priv->nm_device, device);
-
-      default:
-        continue;
+        priv->nm_device = g_object_ref (device);
+        break;
       }
     }
-  }
-
-  if (has_ethernet_connection) {
-    visible = FALSE;
-    goto out;
   }
 
   if (priv->nm_device == NULL) {
--- a/gnome-initial-setup/pages/network/gis-network-page.c
+++ b/gnome-initial-setup/pages/network/gis-network-page.c
@@ -59,6 +59,7 @@ struct _GisNetworkPagePrivate {
   GtkWidget *skip_button;
 
   guint refresh_timeout_id;
+  guint network_handler_id;
 };
 typedef struct _GisNetworkPagePrivate GisNetworkPagePrivate;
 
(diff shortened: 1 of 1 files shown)


gnome-initial-setup/pages/network/gis-network-page.c
```diff
@@ -625,6 +668,24 @@ gis_network_page_constructed (GObject *object)
...
+  /*
+   * Monitor the network to see if the network gets available without
+   * using this panel.
+   */
+  monitor = g_network_monitor_get_default ();
+
+  g_signal_connect (monitor,
+                    "notify::connectivity",
+                    G_CALLBACK (network_connectivity_changed_cb),
+                    page);
```
cosimoc: Should also probably disconnect this signal when the page is destroyed/disposed, since the network monitor is a singleton

gnome-initial-setup/pages/network/gis-network-page.c
```diff
@@ -588,11 +594,48 @@ active_connections_changed (NMClient *client, GParamSpec *pspec, GisNetworkPage
...
+      g_network_monitor_get_connectivity (monitor) != G_NETWORK_CONNECTIVITY_FULL)
+    {
+      return FALSE;
+    }
+
+
+  gtk_widget_hide (GTK_WIDGET (self));
+
+  if (is_current_page)
+    gis_assistant_next_page (gis_driver_get_assistant (GIS_PAGE (self)->driver));
```
cosimoc: I don't think this is right; we never want to change the UI without user interaction. If the connectivity happens to change to full while the page is visible, we should just make the "Next" button active.

gnome-initial-setup/pages/network/gis-network-page.c
```diff
@@ -588,18 +595,53 @@ active_connections_changed (NMClient *client, GParamSpec *pspec, GisNetworkPage
...
+      priv->row_clicked ||
+      g_network_monitor_get_connectivity (monitor) != G_NETWORK_CONNECTIVITY_FULL)
+    {
+      return FALSE;
+    }
+
+  gis_page_set_complete (GIS_PAGE (self), TRUE);
+
+  if (!is_current_page)
+    gtk_widget_hide (GTK_WIDGET (self));
```
cosimoc: This can't really be reached as far as I can see, as the same condition is already checked above.

gnome-initial-setup/pages/network/gis-network-page.c
```diff
@@ -588,11 +595,48 @@ active_connections_changed (NMClient *client, GParamSpec *pspec, GisNetworkPage
...
+  GisAssistant *assistant;
+  gboolean is_current_page;
+
+  priv = gis_network_page_get_instance_private (self);
+  assistant = gis_driver_get_assistant (GIS_PAGE (self)->driver);
+  is_current_page = gis_assistant_get_current_page (assistant) == GIS_PAGE (self);
+
+  if (g_network_monitor_get_connectivity (monitor) != G_NETWORK_CONNECTIVITY_FULL)
+    {
+      if (!is_current_page || priv->row_clicked)
```
wjt: This function does not always return a value. In the case where connectivity != `G_NETWORK_CONNECTIVITY_FULL`, but `is_current_page && !priv->row_clicked`, then what value should be returned?

(GCC should be issuing a warning for this; I don't think it should be part of this branch but we should make upstream g-i-s compile without warnings and then enable -Werror during development.)

gnome-initial-setup/pages/network/gis-network-page.c
```diff
@@ -588,11 +589,47 @@ active_connections_changed (NMClient *client, GParamSpec *pspec, GisNetworkPage
...
+  /*
+   * Hide the Network page if a connection is available and it's not
+   * visible. This way, the user won't be even prompted about network.
+   * Otherwise, make it visible so the user can connect to a network,
+   * keeping in mind this page is only visible with a Wireless adapter.
+   */
+  if (!is_current_page)
+    gtk_widget_set_visible (GTK_WIDGET (self), !has_connection && priv->nm_device);
+
+  gis_page_set_complete (GIS_PAGE (self), has_connection);
```
wjt: There's a really unlikely sequence of events which I think could leave this page in the wrong state:

* User advances to this page
* User ticks the “don't set up my network now” button; `sync_page_complete` is called and sets the page complete
* User plugs in a network cable – this function is called and sets the page complete (again)
* User unplugs the network cable – this function is called again and sets the page incomplete, which is incorrect

Maybe this function and `sync_page_complete` should be combined?

PR #146, opened 2017-01-27
T15278 capitalisation

pwithnall:
Capitalisation and Unicode string fixes for https://phabricator.endlessm.com/T15278.


pwithnall:
Updated version of the branch which updates the `msgid` strings in the `.po` files so that no strings are marked as fuzzy.

pwithnall:
> ui: Apply unicode ellipses to translations

I deliberately didn’t do that because I’m unsure of the semantics of ‘…’ in non-English languages, but if you’re happy with it then I have no complaints. Thanks for reviewing and merging.

PR #148, opened 2017-02-06
T15310

GeorgesStavracas:
https://phabricator.endlessm.com/T15310

### DIFF
--- a/gnome-initial-setup/pages/language/cc-language-chooser.c
+++ b/gnome-initial-setup/pages/language/cc-language-chooser.c
@@ -547,6 +547,7 @@ void
 cc_language_chooser_init (CcLanguageChooser *chooser)
 {
         chooser->priv = GET_PRIVATE (chooser);
+        chooser->priv->showing_extra = TRUE;
 }
 
 void
--- a/gnome-initial-setup/pages/language/cc-language-chooser.c
+++ b/gnome-initial-setup/pages/language/cc-language-chooser.c
@@ -59,6 +59,9 @@ struct _CcLanguageChooserPrivate
         GtkWidget *more_item;
         GtkWidget *filter_entry;
         GtkWidget *language_list;
+        GtkWidget *language_scrolledwindow;
+        GtkWidget *to_be_scrolled_row;
+        gboolean initial_scroll;
         gboolean showing_extra;
         gchar **filter_words;
         gchar *language;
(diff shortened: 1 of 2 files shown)


gnome-initial-setup/pages/language/cc-language-chooser.c
```diff
@@ -145,10 +147,91 @@ language_widget_new (const char *locale_id,
...
+{
+        ScrollData *data;
+
+        data = g_new0 (ScrollData, 1);
+        data->row = row;
+        data->chooser = self;
+
+        /* Nothing is scheduled, we should do it now */
+        if (self->priv->scroll_timeout_id > 0)
+                g_source_remove (self->priv->scroll_timeout_id);
```
cosimoc: It looks to me that the only reason you ever need to schedule a timeout is when the widget is not yet mapped. But you should not use a timeout/retry mechanism to know when the widget gets mapped -- you can simply connect to the `map` signal.

gnome-initial-setup/pages/language/cc-language-chooser.c
```diff
@@ -145,10 +147,91 @@ language_widget_new (const char *locale_id,
...
+
+        vadjustment = gtk_scrolled_window_get_vadjustment (GTK_SCROLLED_WINDOW (self->priv->language_scrolledwindow));
+        page = gtk_adjustment_get_page_size (vadjustment);
+        max = gtk_adjustment_get_upper (vadjustment);
+
+        page_increment = gtk_adjustment_get_page_increment (vadjustment);
+        real_value = get_selected_language_row_y (row) - page_increment / 2.0;
+        value = gtk_adjustment_get_value (vadjustment);
+
+        gtk_adjustment_set_page_increment (vadjustment, real_value - value);
```
cosimoc: This does not seem like the best way to scroll -- the `scroll-child` signal is a keybinding signal that is managed internally by `GtkScrolledWindow`. I think you can just use `gtk_adjustment_set_value()` instead.

GeorgesStavracas: Using `gtk_adjustment_set_value()` is OK, but won't have that nifty scroll animation. This approach was actually suggested by Matthias Clasen himself, so I believe it's OK to use it here.

gnome-initial-setup/pages/language/cc-language-chooser.c
GeorgesStavracas: I'm not sure why, but previous testing showed that using the `map` signal don't work. Not only here, but in other applications as well, nothing happens. Usually, trying to scroll an unmapped scrolledwindow would result in a segfault - but scrolling against the `map` and `map-event` signals does absolutely nothing.

cosimoc: Perhaps because the condition you want to check is whether the widget is realized instead? I am not very comfortable merging the timeout hack without at least understanding why we would be forced to use it.

gnome-initial-setup/pages/language/cc-language-chooser.c
cosimoc: The requirement is really only that it appears centered the first time it comes up pre-selected, not that every click centers it, so I don't think the animation is worth the code. I actually think that it is confusing if it centered itself at every click, since that's not what other scrolled lists do.
So I would
- change the code so that it only centers the initial selection and not further clicks
- remove the animation

gnome-initial-setup/pages/language/cc-language-chooser.c
```diff
@@ -145,10 +148,80 @@ language_widget_new (const char *locale_id,
...
+        ScrollData *data;
+
+        data = g_new0 (ScrollData, 1);
+        data->row = row;
+        data->chooser = self;
+
+        if (gtk_widget_get_realized (self->priv->language_scrolledwindow)) {
+                update_scroll_position (data);
+        } else {
+                self->priv->scroll_handler_id =
```
cosimoc: I don't think you actually need to save this.

gnome-initial-setup/pages/language/cc-language-chooser.c
```diff
@@ -145,10 +148,80 @@ language_widget_new (const char *locale_id,
...
+                 GtkWidget         *row)
+{
+        ScrollData *data;
+
+        data = g_new0 (ScrollData, 1);
+        data->row = row;
+        data->chooser = self;
+
+        if (gtk_widget_get_realized (self->priv->language_scrolledwindow)) {
+                update_scroll_position (data);
```
cosimoc: Nitpick: early return here.

gnome-initial-setup/pages/language/cc-language-chooser.c
```diff
@@ -145,10 +148,80 @@ language_widget_new (const char *locale_id,
...
+
+static void
+schedule_scroll (CcLanguageChooser *self,
+                 GtkWidget         *row)
+{
+        ScrollData *data;
+
+        data = g_new0 (ScrollData, 1);
+        data->row = row;
+        data->chooser = self;
```
cosimoc: At the moment this data struct is leaked; but I think I would prefer it if you just saved the row to scroll to inside the private structure instead of using a separate one.

gnome-initial-setup/pages/language/cc-language-chooser.c
```diff
@@ -145,10 +148,63 @@ language_widget_new (const char *locale_id,
...
+static void
+schedule_scroll (CcLanguageChooser *self,
+                 GtkWidget         *row)
+{
+        self->priv->to_be_scrolled_row = row;
+
+        if (gtk_widget_get_realized (self->priv->language_scrolledwindow)) {
+                update_scroll_position (self);
+                return;
+        } else {
```
cosimoc: Sorry for being so pedantic here, but if we return early, we should also get rid of the `else`

PR #149, opened 2017-02-08
account: implement the new password strength scheme

GeorgesStavracas:
The old password strength validation only starts giving us any
value when the password already has a good strength. This behavior
led to confusion, as it only actually works when after typing a
few random characters, making us think it wasn't working at all.

The new password scheme is just a minor extension of the old one.
Basically, whenever a password is typed, we make sure the strength
to be at least one - instead of 0 like before.

Now, we use the following scheme:

Empty	  -> 0 bars
Weak	  -> 1 bar (red)
Regular	  -> 2 bars (yellow)
Strong	  -> 3 bars (green)
Excellent -> 4 bars (blue)

This makes the feedback much more immediate, and improves the overall
interaction with the initial setup page.

https://phabricator.endlessm.com/T15307

### DIFF
--- a/gnome-initial-setup/pages/account/account.css
+++ b/gnome-initial-setup/pages/account/account.css
@@ -1,0 +1,15 @@
+/* Colors from the Tango palette */
+levelbar trough block.filled.weak {
+    background-color: #ef2929;
+    border-color: #cc0000;
+}
+
+levelbar trough block.filled.regular {
+    background-color: #edd400;
+    border-color: #c4a000;
+}
+
+levelbar trough block.filled.good {
+    background-color: #73d216;
+    border-color: #4e9a06;
+}
--- a/gnome-initial-setup/pages/account/gis-account-page.c
+++ b/gnome-initial-setup/pages/account/gis-account-page.c
@@ -400,6 +400,14 @@ update_password_entries (GisAccountPage *page)
     clear_password_validation_error (page);
 
   pw_strength (password, NULL, username, &hint, &long_hint, &strength_level);
+
+  /*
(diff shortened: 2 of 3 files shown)


gnome-initial-setup/pages/account/gis-account-page.c
```diff
@@ -1218,6 +1226,34 @@ gis_account_page_save_data (GisPage *gis_page)
...
+load_css_overrides (GisAccountPage *self)
+{
+  GtkCssProvider *provider;
+  GError *error;
+  GFile *file;
+
+  error = NULL;
+  provider = gtk_css_provider_new ();
+  file = g_file_new_for_uri ("resource:///org/gnome/initial-setup/account.css");
+  gtk_css_provider_load_from_file (provider, file, &error);
```
cosimoc: You can use `gtk_css_provider_load_from_resource()` and avoid going through `GFile`.

gnome-initial-setup/pages/account/account.css
```diff
@@ -0,0 +1,15 @@
...
+
+levelbar trough block.filled.regular {
+    background-color: #edd400;
+    border-color: #c4a000;
+}
+
+levelbar trough block.filled.good {
+    background-color: #73d216;
+    border-color: #4e9a06;
+}
```
cosimoc: Does something like this not come as part of the default theme?

GeorgesStavracas: No

cosimoc: Pretty sure that at least some (warning/success) do... Perhaps not all three levels though. I am OK to have the custom theming here in any case.

PR #151, opened 2017-02-16
live: don't clobber image-specific icon grid; restore Chromium

wjt:
https://phabricator.endlessm.com/T15370
https://phabricator.endlessm.com/T15438

### DIFF
--- a/gnome-initial-setup/pages/live-chooser/eos-setup-live-user
+++ b/gnome-initial-setup/pages/live-chooser/eos-setup-live-user
@@ -17,6 +17,7 @@
 # 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 
 import argparse
+import glob
 import json
 import logging
 import os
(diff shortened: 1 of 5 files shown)


gnome-initial-setup/pages/live-chooser/eos-setup-live-user
```diff
@@ -212,6 +174,54 @@ class SetupLiveUser(object):
...
+
+                with open(path, 'w') as f:
+                    f.write(grid)
+        except:
+            log.exception("while processing %s", path)
+
+
+def prepend_installer_to_icon_grid():
+    log.info("Adding installer icon to the desktop")
+    path = os.path.join(ICON_GRID_DIR, 'icon-grid-prepend-C.json')
```
rshuler: I think there is a case where this won't quite work right...  If the image builder already specifies an icon-grid-prepend-<locale>.json, where <locale> is something other than C, that localized file will be read rather than the C, so your prepend will be ignored.

rshuler: I mean that the shell will read the localized version rather than the one that you modified.

PR #155, opened 2017-03-09
Allow defining a branding-enabled welcome page

mariospr:
This pull request adds a new page "branding-welcome" that will allow us defining the relevant parameters (title, description, logo) for a branded welcome page, in an optional way, as well as an optional list of pages we might want to skip.

To do this we will read a "key file" in `/etc/eos-branding/gnome-initial-setup/pages.conf` that would define the following parameters:
  * `title` (required): string containing a short string to show as title.
  * `description` (optional): string containing long text, likely to be wrapped.
  * `logo` (optional): absolute path to the file with a logo for the brand.
  * `skip` (optional): comma-separated list of PAGE IDs the branded g-i-s should skip.

For example, this is how this file would look on a branded image defining a title,
a description, a logo and skipping the "language" and "keyboard" pages.
```
[Welcome]
title=A title to be shown at the top
description=A long description that will be shown at the bottom of this page, right below the branded logo (if any), explaining what the branded edition is about.
logo=/path/to/the/image/with/the/logo.png
skip=language,keyboard
```

Additionally, this PR refactors a bit the code around GisPage and the language page, so that we can now open the factory mode dialog from any page (not just the language one), which will be useful for the new "branding-welcome" page, which also needs that functionality.

https://phabricator.endlessm.com/T15834


gnome-initial-setup/pages/branding-welcome/gis-branding-welcome-page.c
```diff
@@ -0,0 +1,289 @@
...
+
+static void
+gis_branding_welcome_page_init (GisBrandingWelcomePage *page)
+{
+  GisBrandingWelcomePagePrivate *priv = gis_branding_welcome_page_get_instance_private (page);
+
+  priv->title = NULL;
+  priv->description = NULL;
+  priv->logo_path = NULL;
+  priv->skip_pages = NULL;
```
cosimoc: There's no need to initialize these to NULL -- GObject will do that for the whole private structure automatically.

gnome-initial-setup/pages/branding-welcome/gis-branding-welcome-page.c
```diff
@@ -0,0 +1,289 @@
...
+
+  if (priv->skip_pages != NULL) {
+    g_auto(GStrv) skip_pages = g_strsplit (priv->skip_pages, ",", -1);
+    gint i;
+
+    for (i = 0; skip_pages[i]; i++) {
+      g_debug ("Skipping page %s...", skip_pages[i]);
+      gis_driver_add_skipped_page (driver, skip_pages[i]);
+    }
+  }
```
cosimoc: Could this not be done from the constructed implementation, after the configuration has been loaded?

gnome-initial-setup/pages/branding-welcome/gis-branding-welcome-page.c
```diff
@@ -0,0 +1,289 @@
...
+  attr_list = pango_attr_list_new ();
+
+  /* Rising the baseline of the 14px font (see branding-welcome-page.css)
+   * by 6px gives us a 20px line height. Note that there should be a way
+   * to get this value programatically, but all the experiments I did with
+   * Pango failed so far for some reason, so I'm hardcoding this for now. */
+  attr = pango_attr_rise_new (6 * PANGO_SCALE);
+
+  pango_attr_list_insert (attr_list,attr);
+  gtk_label_set_attributes (GTK_LABEL (WID ("branding-text")), attr_list);
```
cosimoc: You should be able to do this from the builder XML; see https://git.gnome.org/browse/gtk+/tree/gtk/gtklabel.c#n106

gnome-initial-setup/pages/branding-welcome/gis-branding-welcome-page.c
```diff
@@ -0,0 +1,289 @@
...
+  GisBrandingWelcomePagePrivate *priv = NULL;
+  g_autoptr(GKeyFile) keyfile = NULL;
+  g_autoptr(GError) error = NULL;
+  g_auto(GStrv) groups = NULL;
+  gboolean ret = FALSE;
+  gsize n_groups;
+  int i;
+
+  if (!g_file_test (CONFIG_FILEPATH, G_FILE_TEST_EXISTS))
+    goto out;
```
cosimoc: There's no need to check for this  separately -- `g_key_file_load_from_file()` will fail when it does not exist. You can check the error if you don't want to warn in that case.

gnome-initial-setup/pages/branding-welcome/gis-branding-welcome-page.c
```diff
@@ -0,0 +1,289 @@
...
+
+  priv = gis_branding_welcome_page_get_instance_private (page);
+  groups = g_key_file_get_groups (keyfile, &n_groups);
+  for (i = 0; i < n_groups; i++)
+    {
+      size_t len;
+
+      /* We only support "Welcome" branded pages for now */
+      if (g_strcmp0 (groups[i], "Welcome") != 0)
+        continue;
```
cosimoc: Instead of looping over the groups, just call `g_key_file_get_string()` with `Welcome` as the group.

gnome-initial-setup/pages/branding-welcome/gis-branding-welcome-page.c
```diff
@@ -0,0 +1,289 @@
...
+  int i;
+
+  if (!g_file_test (CONFIG_FILEPATH, G_FILE_TEST_EXISTS))
+    goto out;
+
+  keyfile = g_key_file_new ();
+  if (!g_key_file_load_from_file (keyfile, CONFIG_FILEPATH, G_KEY_FILE_NONE, &error))
+    {
+      g_warning ("Could not read file %s: %s\n", CONFIG_FILEPATH, error->message);
+      goto out;
```
cosimoc: Since there's nothing to free at the end (you use auto pointers), return immediately here.

gnome-initial-setup/pages/branding-welcome/gis-branding-welcome-page.c
```diff
@@ -0,0 +1,289 @@
...
+static void
+load_css_overrides (GisBrandingWelcomePage *page)
+{
+  g_autoptr(GtkCssProvider) provider;
+  g_autoptr(GFile) file;
+  g_autoptr(GError) error = NULL;
+
+  provider = gtk_css_provider_new ();
+  file = g_file_new_for_uri ("resource:///org/gnome/initial-setup/branding-welcome-page.css");
+  gtk_css_provider_load_from_file (provider, file, &error);
```
cosimoc: Use `gtk_css_provider_load_from_resource()`.

gnome-initial-setup/pages/branding-welcome/gis-branding-welcome-page.c
```diff
@@ -0,0 +1,289 @@
...
+ * For example, this is how this file would look on a branded image defining a title,
+ * a description, a logo and skipping the "language" and "keyboard" pages.
+ *
+ *   [Welcome]
+ *   title=A title to be shown at the top
+ *   description=A long description that will be shown at the bottom of this page, right below the branded logo (if any), explaining what the branded edition is about.
+ *   logo=/path/to/the/image/with/the/logo.png
+ *   skip=language,keyboard
+ */
+#define CONFIG_FILEPATH CONFIGDIR "/eos-branding/gnome-initial-setup/pages.conf"
```
cosimoc: I would put this in a directory that belongs to g-i-s, not `eos-branding`. Also I'm not sure that we need to make this a generic "pages" configuration. Instead, you could just add a configuration file for the branding page.

gnome-initial-setup/Makefile.am
```diff
@@ -11,17 +11,27 @@ AM_CPPFLAGS = \
...
-gis-assistant-resources.c: gis-assistant.gresource.xml $(resource_files)
+assistant_resource_files = $(shell glib-compile-resources --sourcedir=$(srcdir) --generate-dependencies $(srcdir)/gis-assistant.gresource.xml)
+gis-assistant-resources.c: gis-assistant.gresource.xml $(assistant_resource_files)
 	$(AM_V_GEN) $(GLIB_COMPILE_RESOURCES) --target=$@ --sourcedir=$(srcdir) --generate-source $<
-gis-assistant-resources.h: gis-assistant.gresource.xml $(resource_files)
+gis-assistant-resources.h: gis-assistant.gresource.xml $(assistant_resource_files)
 	$(AM_V_GEN) $(GLIB_COMPILE_RESOURCES) --target=$@ --sourcedir=$(srcdir) --generate-header $<
 BUILT_SOURCES += gis-assistant-resources.c gis-assistant-resources.h
 
+page_resource_files = $(shell glib-compile-resources --sourcedir=$(srcdir) --generate-dependencies $(srcdir)/gis-page.gresource.xml)
```
cosimoc: Instead of moving everything to `GisPage`, I'd rather you moved it to its own file.

gnome-initial-setup/pages/branding-welcome/gis-branding-welcome-page.c
```diff
@@ -214,6 +226,12 @@ gis_branding_welcome_page_constructed (GObject *object)
   update_line_height_for_description (page);
   load_css_overrides (page);
 
+  /* Use ctrl+f to show factory dialog */
+  priv->accel_group = gtk_accel_group_new ();
+  closure = g_cclosure_new_swap (G_CALLBACK (gis_page_show_factory_dialog), page, NULL);
+  gtk_accel_group_connect (priv->accel_group, GDK_KEY_f, GDK_CONTROL_MASK, 0, closure);
+  g_closure_unref (closure);
```
cosimoc: The way this is done typically is by adding a `GAction` and `gtk_application_set_accels_for_action()`. I won't block the PR if you don't want to bother though.

gnome-initial-setup/gis-driver.c
```diff
@@ -127,6 +128,9 @@ gis_driver_finalize (GObject *object)
   g_free (priv->personality);
   g_free (priv->lang_id);
 
+  g_list_foreach (priv->skip_pages, (GFunc)g_free, NULL);
+  g_list_free (priv->skip_pages);
```
mariospr: Ok

gnome-initial-setup/gis-driver.c
```diff
@@ -204,6 +208,32 @@ gis_driver_get_user_language (GisDriver *driver)
...
+
+gboolean
+gis_driver_is_skipped_page (GisDriver *driver, const gchar *page_id)
+{
+  GisDriverPrivate *priv = gis_driver_get_instance_private (driver);
+  GList *item;
+
+  for (item = priv->skip_pages; item; item = g_list_next (item)) {
+    if (g_strcmp0 ((char *)item->data, page_id) == 0)
+      return TRUE;
```
mariospr: It would provide direct access, sure, but this list is likely not going to be long anyway (typically one item), so I'm not sure it's worth. No strong opinions, though.

gnome-initial-setup/pages/account/gis-account-page.c
```diff
@@ -1403,6 +1403,9 @@ gis_account_page_init (GisAccountPage *page)
 void
 gis_prepare_account_page (GisDriver *driver)
 {
+  if (gis_driver_is_skipped_page (driver, PAGE_ID))
+    return;
```
mariospr: I'm not entirely sure that will work here, but I will give it a try. Thanks

gnome-initial-setup/pages/branding-welcome/gis-branding-welcome-page.c
mariospr: I'm just following the lead of the language page here, and I'd rather leave it like this for now if it's ok, to avoid even more refactoring.

gnome-initial-setup/pages/branding-welcome/gis-branding-welcome-page.c
mariospr: Actually, my initial approach put everything in /etc/gnome-initial-setup, but then I started to work on https://phabricator.endlessm.com/T15836 and, when I was about to do something similar, creating /etc/gnome-control-center, I realized both are "2 faces of the same coin" and that probably looking in a common directory for branding made sense, not to scatter too many custom directories in /etc.

This way, whis is how /etc would look in my VM after having the work for T15834 and T15836 implemented there:
```
/etc/eos-branding/
/etc/eos-branding/assets
/etc/eos-branding/assets/logo-control-center.svg
/etc/eos-branding/assets/logo-initial-setup.png
/etc/eos-branding/gnome-control-center
/etc/eos-branding/gnome-control-center/info.conf
/etc/eos-branding/gnome-initial-setup
/etc/eos-branding/gnome-initial-setup/pages.conf
```

Does that makes more sense now?

gnome-initial-setup/pages/branding-welcome/gis-branding-welcome-page.c
mariospr: Followed the lead of the summary page here, but sure. I'll change that.

gnome-initial-setup/pages/branding-welcome/gis-branding-welcome-page.c
mariospr: ok.

gnome-initial-setup/pages/branding-welcome/gis-branding-welcome-page.c
mariospr: True

gnome-initial-setup/pages/branding-welcome/gis-branding-welcome-page.c
mariospr: Sure

gnome-initial-setup/pages/branding-welcome/gis-branding-welcome-page.c
mariospr: I looked around for a while and I haven't really found a way to do this that is not using Pango.

How exactly would you do it from the builder XML file?

gnome-initial-setup/pages/branding-welcome/gis-branding-welcome-page.c
mariospr: I always hesitate about these things, thanks for clarifying.

gnome-initial-setup/pages/branding-welcome/gis-branding-welcome-page.c
mariospr: Sounds good. I'll do

gnome-initial-setup/Makefile.am
mariospr: I can create a gis-util or something like that, ok

gnome-initial-setup/pages/branding-welcome/gis-branding-welcome-page.c
mariospr: Actually, I think I'd rather have this if you don't mind, which is what's used everywhere else in g-i-s and also because that way I can easily check for errors using GError,  Using `gtk_css_provider_load_from_resource()` would force me to connect to the "parsing-error" signal which is a bit overkill and inconvenient here.

gnome-initial-setup/pages/account/gis-account-page.c
mariospr: The problem with the skip file in the driver is that it gets considered **before** the GisPage instances are created, which is before we give the branding-welcome page a chance to load the data from the configuration file, and parse it, so I'd rather not use that.

That said, I did it this way because I wanted to prevent the page from being added to the driver at all, but if hiding it in the constructor works too that's fine as well for me.

mariospr: Hmm.. I tried that, but it does not work as expected: while it does skip the desired page (language, in my case) when clicking on "Next", the "dots indicator" at the bottom does still reflect that there is something in there, because I can see how it moves from the first dot to the third one when skipping over the page I don't want to show.

I think it might be better to leave it as I have it now, then...

gnome-initial-setup/gis-driver.c
cosimoc: There should be no additional difficulty in using a `GHashTable`, and if it's one item only, I'd rather you actually just made it a `char *` instead...

gnome-initial-setup/pages/branding-welcome/gis-branding-welcome-page.c
cosimoc: Please change it; that function does not return an error because it will always succeed in the cases where the other would have failed, which is when the file is missing.
CSS parsing errors are not reported through `GError`.

gnome-initial-setup/pages/branding-welcome/gis-branding-welcome-page.c
cosimoc: Something like
```
<attributes>
  <attribute name="rise" value="6"/>
</attributes>
```

gnome-initial-setup/pages/branding-welcome/gis-branding-welcome-page.c
cosimoc: I did not realize that this code is not new and only being moved from another page. If that's the case, I'm OK keeping it as-is.

gnome-initial-setup/pages/account/gis-account-page.c
cosimoc: That sounds like a bug in the dots indicator, but my more general point is that the same code that calls `gjs_prepare_*_page()` could do that check, instead of sprinkling it in every page.
It also feels wrong to me that the branding page dictates behavior of the other pages; pages should be independent of each other.

An alternative approach that I prefer would be to remove the skip list from the branding keyfile, and ship the regular skip file together with the branding page keyfile configuration.

gnome-initial-setup/gis-driver.c
mariospr: Ok, Hash table it is then

gnome-initial-setup/pages/branding-welcome/gis-branding-welcome-page.c
mariospr: Ok.

gnome-initial-setup/pages/branding-welcome/gis-branding-welcome-page.c
mariospr: Interesting, I did not know that. Thanks!

gnome-initial-setup/pages/branding-welcome/gis-branding-welcome-page.c
cosimoc: The reason I am not a huge fan of it is because it makes it impossible to upstream it; I think that if we used different paths for the customization point (or if we made the path for the customization point itself also configurable e.g. through a build option) we could make a case for these to also go upstream.

However if you're short on time, I'd be OK accepting this as-is.

gnome-initial-setup/pages/account/gis-account-page.c
mariospr: I re-checked the skip file approach and I actually like it more. I was a bit blinded with providing the list of skip files from the branding file itself, but we can actually ship a separate skip file using the logic that is already in place and that would remove all the other code that I added.

gnome-initial-setup/pages/branding-welcome/gis-branding-welcome-page.c
mariospr: I think I just changed my mind: I think we can just do /etc/gnome-initial-setup here and that's it.

gnome-initial-setup/pages/branding-welcome/gis-branding-welcome-page.c
```diff
@@ -0,0 +1,212 @@
...
+                                             GTK_STYLE_PROVIDER_PRIORITY_APPLICATION);
+}
+
+static gboolean
+read_config_file (GisBrandingWelcomePage *page)
+{
+  GisBrandingWelcomePagePrivate *priv = NULL;
+  g_autoptr(GKeyFile) keyfile = NULL;
+  g_autoptr(GError) error = NULL;
+  g_auto(GStrv) groups = NULL;
```
cosimoc: This is unused

gnome-initial-setup/pages/branding-welcome/gis-branding-welcome-page.c
```diff
@@ -0,0 +1,212 @@
...
+    g_debug ("No branding configuration found");
+    return;
+  }
+
+  gtk_label_set_label (GTK_LABEL (WID("branding-title")), priv->title);
+
+  if (priv->description != NULL) {
+    opt_widget = WID("branding-text");
+    gtk_label_set_label (GTK_LABEL (opt_widget), priv->description);
+    gtk_widget_set_visible (opt_widget, TRUE);
```
cosimoc: Nitpick: use `gtk_widget_show()` for consistency (also below)

gnome-initial-setup/pages/branding-welcome/gis-branding-welcome-page.c
mariospr: Oops! I thought I removed it, but clearly not

gnome-initial-setup/pages/branding-welcome/gis-branding-welcome-page.c
mariospr: Ok

PR #175, opened 2017-06-15
summary: limit width of live mode warning

wjt:
Unscientific straw polls of the London team concluded that this is about the right point to wrap the text, and that a little extra margin below makes the elements look more balanced.

https://phabricator.endlessm.com/T17622


cosimoc:
>Unscientific straw polls of the London team
Good enough for me :-)

PR #178, opened 2017-06-23
display: update bottom-right corner asset

wjt:
This is a little wider than the old asset to allow space for a bit of empty taskbar on the left, which I think looks better. (The icons are wider in the new design, partly because the power status is shown here.)

https://phabricator.endlessm.com/T17471


wjt:
screenshot from 2017-06-23 06-58-34

wjt:
Updated to show a charging symbol:

screenshot from 2017-06-23 15-08-23

I also have an image prepared to show the battery discharging but at full capacity:

screenshot from 2017-06-23 15-06-32

PR #181, opened 2017-06-29
set the envvar LC_MESSAGES

rshuler:
When the user sets their preferred language, the environment variable
LC_MESSAGES shall be set too, because gis might spawns processes (e.g.
gkbd-keyboard-display) and they should be localised accordingly.

https://phabricator.endlessm.com/T3172
https://phabricator.endlessm.com/T17967


mcatanzaro:
Closing per https://phabricator.endlessm.com/T17967#415101

PR #189, opened 2017-07-24
timezone: Fix displaying the right clock format

joaquimrocha:
When the user changes the locale, then the clock format in the timezone
page should reflect that, but it was apparently always using the initial
(global) locale.

These changes make sure that we get the clock format for the currently
set locale, which will honor what the user has previously chosen inside
g-i-s.

https://phabricator.endlessm.com/T17468

### DIFF
--- a/gnome-initial-setup/pages/timezone/gis-timezone-page.c
+++ b/gnome-initial-setup/pages/timezone/gis-timezone-page.c
@@ -32,6 +32,7 @@
 #include <stdlib.h>
 #include <string.h>
 #include <langinfo.h>
+#include <locale.h>
 
 #define GNOME_DESKTOP_USE_UNSTABLE_API
 #include <libgnome-desktop/gnome-languages.h>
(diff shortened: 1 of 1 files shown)


gnome-initial-setup/pages/timezone/gis-timezone-page.c
```diff
@@ -78,13 +79,17 @@ typedef struct _GisTimezonePagePrivate GisTimezonePagePrivate;
...
 static GDesktopClockFormat
 get_default_time_format (void)
 {
   const char *ampm, *nl_fmt;
+  const char *locale_str = setlocale (LC_ALL, NULL);
+  g_auto(locale_t) locale = newlocale (LC_ALL_MASK, locale_str, NULL);
 
   /* Default to 24 hour if we can't get the format from the locale */
-  nl_fmt = nl_langinfo (T_FMT);
+  nl_fmt = nl_langinfo_l (T_FMT, locale);
```
cosimoc: Does the other call below to `nl_langinfo()` not require the same treatment?

joaquimrocha: @cosimoc I guess that's right, though I didn't see it failing without it. Ihave just updated the PR.

PR #193, opened 2017-08-17
language: allow to specify initial languages from vendor.conf

cosimoc:
Instead of hardcoding the initial list, we can use vendor.conf to seed
it; we still use the hardcoded defaults when that's not the case.

https://phabricator.endlessm.com/T17467

### DIFF
--- a/gnome-initial-setup/pages/language/cc-common-language.c
+++ b/gnome-initial-setup/pages/language/cc-common-language.c
@@ -285,13 +285,57 @@ insert_user_languages (GHashTable *ht)
         }
 }
 
-GHashTable *
-cc_common_language_get_initial_languages (void)
+#define VENDOR_LANGUAGE_GROUP "Language"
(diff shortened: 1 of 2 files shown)


gnome-initial-setup/pages/language/cc-common-language.c
```diff
@@ -285,13 +285,59 @@ insert_user_languages (GHashTable *ht)
...
+#define VENDOR_LANGUAGE_GROUP "Language"
+#define VENDOR_LANGUAGE_INITIAL_LANGUAGES_KEY "initial_languages"
+
+static gboolean
+insert_vendor_languages (GHashTable *ht)
 {
-        GHashTable *ht;
+        g_autoptr(GKeyFile) keyfile = NULL;
+        g_autoptr(GError) error = NULL;
+        g_autofree char **languages = NULL;
```
ramcq: I think this is the wrong free macro, shouldn't we have g_autoptr(GStrv)? g_key_file_get_string_list is transfer_full and insert_language copies the language it's given.

PR #194, opened 2017-08-23
live-chooser: refer to DVD, not USB, if booted from DVD

wjt:
https://phabricator.endlessm.com/T18708


gnome-initial-setup/gis-driver.c
```diff
@@ -87,33 +87,55 @@ typedef struct _GisDriverPrivate GisDriverPrivate;
 
 G_DEFINE_TYPE_WITH_PRIVATE(GisDriver, gis_driver, GTK_TYPE_APPLICATION)
 
+/* Should be kept in sync with eos-installer */
 static gboolean
-running_live_session (void)
+check_for_live_boot (gchar **uuid)
```
cosimoc: I'd prefer it if this parameter was optional; alternatively you could have two functions, one that detects live boots, and another one that detects the UUID. That way you could pass NULL (or avoid getting the UUID altogether) if desired.

gnome-initial-setup/gis-driver.c
```diff
@@ -127,14 +132,67 @@ check_for_live_boot (gchar **uuid)
   return live_boot;
 }
 
+static void
+udisks_client_new_cb (
+    GObject *source,
```
cosimoc: Nitpick: move this on previous line

gnome-initial-setup/gis-driver.c
```diff
@@ -127,14 +132,67 @@ check_for_live_boot (gchar **uuid)
...
+       */
+      for (l = blocks; l != NULL && !priv->is_live_dvd; l = l->next)
+        {
+          UDisksBlock *block = UDISKS_BLOCK (l->data);
+          UDisksDrive *drive = udisks_client_get_drive_for_block (udisks, block);
+
+          if (drive != NULL && udisks_drive_get_optical (drive))
+            {
+              priv->is_live_dvd = TRUE;
+              g_object_notify_by_pspec (G_OBJECT (driver), obj_props[PROP_LIVE_DVD]);
```
cosimoc: It probably does not matter in practice, since the driver is always alive, but I find it a bit surprising that this could potentially notify after the page is "morally" defunct.
Instead you could attach a cancellable to the async operation and return here when that's the case.

gnome-initial-setup/gis-driver.c
wjt: This parameter is not optional because the function is listed verbatim from `eos-installer` where the UUID is always required. It is also always required here, so I don't see the advantage of making it optional!

gnome-initial-setup/gis-driver.c
wjt: I don't really understand what you mean here – by "morally defunct" do you mean "when the driver has no external references"? In that case, it should also have no external signal listeners. As you say, the driver is a singleton which survives the length of the process.

wjt: Asked another way: who would cancel the cancellable? Are you proposing that the `udisks_client_new()` call would not own a ref to the `GisDriver` and the driver would cancel an internal cancellable when disposed?

gnome-initial-setup/gis-driver.c
wjt: Fixed.

gnome-initial-setup/gis-driver.c
cosimoc: Oh I see... I was confused by the fact that it seemed optional from the first commit. In that case perhaps you want to assert that `uuid != NULL`?

gnome-initial-setup/gis-driver.c
cosimoc: "morally defunct" = the only reference is the one held by the asynchronous initialization of the udisks client.
Yes indeed, I am proposing to pass a cancellable to the operation and cancel it when the driver is disposed. (Even though we both agree that it does not make a lot of difference, since the driver is a process-wide singleton.)

gnome-initial-setup/gis-driver.c
wjt: I have added `g_return_val_if_fail (uuid != NULL, FALSE);`.

gnome-initial-setup/gis-driver.c
wjt: Fine. I have added a patch to cancel the cancellable when the GisDriver is shut down. (I think you were proposing to not take a ref for the `udisk_client_new()` call but I really dislike this pattern of relying on the precise semantics of whether the returned error is CANCELLED for whether or not the user_data pointer is valid or not.) As explained in the commit message, the process actually never terminates cleanly except in the reformatter path anyway so this code will never run, but okay.

While looking for calls to `g_application_quit`, and while testing this change, I discovered that the error handling path for running `eos-installer` did not work, so I have fixed that. I also fixed some compiler warnings introduced in the most recent rebase.

gnome-initial-setup/gis-driver.c
```diff
@@ -794,6 +797,17 @@ gis_driver_startup (GApplication *app)
 }
 
 static void
+gis_driver_shutdown (GApplication *app)
+{
+  GisDriver *driver = GIS_DRIVER (app);
+  GisDriverPrivate *priv = gis_driver_get_instance_private (driver);
+
+  g_cancellable_cancel (priv->cancellable);
```
cosimoc: The cancellable is leaked here

gnome-initial-setup/gis-driver.c
```diff
@@ -170,7 +173,7 @@ udisks_client_new_cb (GObject      *source,
 
       g_list_free_full (blocks, g_object_unref);
     }
-  else
+  else if (!g_cancellable_is_cancelled (priv->cancellable))
```
cosimoc: Optional: you could check here if the error matches the cancelled code instead of looking directly at the cancellable state.

gnome-initial-setup/gis-driver.c
wjt: It's now unref'd in `finalize()`

gnome-initial-setup/gis-driver.c
wjt: Ok

PR #195, opened 2017-08-24
language: make sure initial language is always at the top

cosimoc:
It could be confusing to have to scroll back to see other languages;
make sure the initial language is always the first in the list.

https://phabricator.endlessm.com/T17467

### DIFF
--- a/gnome-initial-setup/pages/language/cc-language-chooser.c
+++ b/gnome-initial-setup/pages/language/cc-language-chooser.c
@@ -52,6 +52,7 @@ struct _CcLanguageChooserPrivate
 
         gboolean showing_extra;
         gchar *language;
+        gchar *initial_language;
 };
 typedef struct _CcLanguageChooserPrivate CcLanguageChooserPrivate;
 G_DEFINE_TYPE_WITH_PRIVATE (CcLanguageChooser, cc_language_chooser, GTK_TYPE_BOX);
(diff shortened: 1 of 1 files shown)


gnome-initial-setup/pages/language/cc-language-chooser.c
```diff
@@ -569,6 +578,7 @@ cc_language_chooser_constructed (GObject *object)
 
         if (priv->language == NULL)
                 priv->language = cc_common_language_get_current_language ();
+        priv->initial_language = g_strdup (priv->language);
 
         sync_all_checkmarks (chooser);
         show_more (chooser);
```
rshuler: Not in the new code added, but I think the return value of an int function is undefined if not specified.

gnome-initial-setup/pages/language/cc-language-chooser.c
```diff
@@ -415,6 +416,8 @@ sort_languages (GtkListBoxRow *a,
                 GtkListBoxRow *b,
                 gpointer       data)
 {
+        CcLanguageChooser *chooser = data;
+        CcLanguageChooserPrivate *priv = cc_language_chooser_get_instance_private (chooser);
         LanguageWidget *la, *lb;
         gchar *normalized_a, *normalized_b;
         gint retval;
```
rshuler: Also not in the new code, but it appears that the 'retval' variable is never used.

gnome-initial-setup/pages/language/cc-language-chooser.c
rshuler: Sorry -- ignore that.  I missed expanding part of the code and mixed two functions together.

gnome-initial-setup/pages/language/cc-language-chooser.c
rshuler: Ignore that, too.  Again, my error in not expanding all the code when reviewing in github.

PR #198, opened 2017-08-28
demo mode: hook up cancel button to hide the dialog

erikos:
This is a follow up for https://phabricator.endlessm.com/T18734

### DIFF
--- a/gnome-initial-setup/gis-page-util.c
+++ b/gnome-initial-setup/gis-page-util.c
@@ -448,12 +448,15 @@ gis_page_util_show_factory_dialog (GisPage *page)
 }
 
 static void
-pressed_demo_button (GtkWidget *widget, gpointer user_data)
+response_cb (GtkDialog *dialog,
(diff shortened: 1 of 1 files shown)


gnome-initial-setup/gis-page-util.c
```diff
@@ -448,20 +448,27 @@ gis_page_util_show_factory_dialog (GisPage *page)
...
-pressed_demo_button (GtkWidget *widget, gpointer user_data)
+response_cb (GtkDialog *dialog,
+             guint      response_id,
+             gpointer   user_data)
 {
   GisDriver *driver = GIS_DRIVER (user_data);
 
+  if (response_id == GTK_RESPONSE_CANCEL)
+  {
+    gtk_widget_hide (GTK_WIDGET (dialog));
```
smspillaz: We're using `gtk_window_close` on the dialog below, we should either change the latter or make this consistent with it.

PR #202, opened 2017-08-31
T18995: Indicate visually that we are in demo mode

smspillaz:
https://phabricator.endlessm.com/T18995


gnome-initial-setup/gis-driver.c
```diff
@@ -52,6 +52,7 @@ gis_driver_mode_get_type (void) {
 enum {
   REBUILD_PAGES,
   LOCALE_CHANGED,
+  ENTER_DEMO_MODE,
```
erikos: I would name the signal DEMO_MODE_ENTERED or ENTERED_DEMO_MODE.

PR #204, opened 2017-09-01
page-util: don't enter demo mode for live sessions

cosimoc:
https://phabricator.endlessm.com/T19032


smspillaz:
I guess this works for now, though I thought we would have just had a gsetting set by the image builder so that we could disable this on an image-by-image basis.

PR #207, opened 2017-09-04
T18983

smspillaz:
Reset the machine id and record a metric when we enter demo mode. Metrics will be set to enabled by default.

https://phabricator.endlessm.com/T18983

### DIFF
--- a/gnome-initial-setup/gis-driver.c
+++ b/gnome-initial-setup/gis-driver.c
@@ -420,6 +420,37 @@ create_demo_user (GisDriver *driver, GError **error)
   return TRUE;
 }
 
+static gboolean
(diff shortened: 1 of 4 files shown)


gnome-initial-setup/gis-driver.c
```diff
@@ -420,6 +420,40 @@ create_demo_user (GisDriver *driver, GError **error)
...
+                               G_DBUS_CALL_FLAGS_NONE,
+                               -1,
+                               NULL,
+                               error);
+
+  if (!rv)
+    return FALSE;
+
+  g_variant_unref (rv);
+  g_object_unref (proxy);
```
wjt: Use `g_autoptr()` and avoid these two.

PR #219, opened 2017-09-12
Demo mode: use as well the default avatar

erikos:
Use the same default avatar (the sunflower)
as for any other EOS user.

https://phabricator.endlessm.com/T19144

### DIFF
--- a/gnome-initial-setup/gis-driver.c
+++ b/gnome-initial-setup/gis-driver.c
@@ -407,6 +407,7 @@ create_demo_user (GisDriver *driver, GError **error)
 
   act_user_set_password_mode (user, ACT_USER_PASSWORD_MODE_NONE);
   act_user_set_automatic_login (user, TRUE);
+  act_user_set_icon_file (user, AVATAR_IMAGE_DEFAULT);
 
   language = gis_driver_get_user_language (driver);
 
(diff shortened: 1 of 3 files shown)


gnome-initial-setup/gis-driver.c
```diff
@@ -370,6 +370,7 @@ gis_driver_is_in_demo_mode (GisDriver *driver)
 
 #define DEMO_ACCOUNT_USERNAME "demo-guest"
 #define DEMO_ACCOUNT_FULLNAME "Demo"
+#define DEMO_ACCOUNT_AVATAR_IMAGE "/usr/share/pixmaps/faces/sunflower.jpg"
```
smspillaz: Is this not defined somewhere else? I would imagine it is done for regular users too?

smspillaz: Would it make sense to share the define?

erikos: I guess we could move the define to "gnome-initial-setup/gnome-initial-setup.h" since that is included in both places.

gnome-initial-setup/gnome-initial-setup.h
```diff
@@ -38,5 +38,7 @@ typedef struct _GisPage      GisPage;
 
 void gis_ensure_stamp_files (void);
 
+#define AVATAR_IMAGE_DEFAULT "/usr/share/pixmaps/faces/sunflower.jpg"
```
smspillaz: You probably want to define this at the top of the file after the `#include`s

smspillaz:
Looks good to me - thanks for picking this up!

PR #223, opened 2017-12-07
passwordPage: make latin kbd layout (en) available for password entries

erikos:
If the current keyboard layout is not a latin one we will add the fallback kbd layout 'en'. This calls out to gnome-shell where the logic to detect the keyboard layout is kept. The layout will be available in the popup switcher and the taskbar keyboard layout switcher.

https://phabricator.endlessm.com/T18598

### DIFF
--- a/gnome-initial-setup/pages/password/gis-password-page.c
+++ b/gnome-initial-setup/pages/password/gis-password-page.c
@@ -47,6 +47,9 @@ struct _GisPasswordPagePrivate
   GtkWidget *confirm_explanation;
   GtkWidget *password_toggle;
   GtkWidget *reminder_entry;
+
+  GDBusProxy *input_source_manager;
+
   gboolean valid_confirm;
   gboolean valid_password;
   guint timeout_id;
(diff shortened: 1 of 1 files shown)


gnome-initial-setup/pages/password/gis-password-page.c
```diff
@@ -320,6 +369,18 @@ gis_password_page_constructed (GObject *object)
 
   validate (page);
 
+  priv->cancellable = g_cancellable_new ();
+
+  g_dbus_proxy_new_for_bus (G_BUS_TYPE_SESSION,
```
cosimoc: See similar comments about it being better to build this synchronously in the g-c-c PR

gnome-initial-setup/pages/password/gis-password-page.c
```diff
@@ -320,6 +347,14 @@ gis_password_page_constructed (GObject *object)
 
   validate (page);
 
+  priv->input_source_manager = g_dbus_proxy_new_for_bus_sync (G_BUS_TYPE_SESSION,
```
cosimoc: It occurred to me that we don't actually need to keep this object around, and we could instead use `g_dbus_connection_call_sync()` when setting the property, but this is fine too.

gnome-initial-setup/pages/password/gis-password-page.c
```diff
@@ -127,12 +130,36 @@ gis_password_page_save_data (GisPage *gis_page)
   gis_update_login_keyring_password (password);
 }
 
+void
```
cosimoc: Should be `static`... I will push a small fixup.

PR #226, opened 2018-02-13
Deduplicate live and demo mode setup

wjt:
This PR started as an attempt to reduce the overlap between `eos-setup-live-user` and `gnome-initial-setup-copy-worker`, by moving all of the former's functions to a system-wide service launched by systemd. I then realised that entering demo mode performs many of the same configuration changes, so unified the two.

The top commit is filed upstream [here](https://bugzilla.gnome.org/show_bug.cgi?id=793429) so it would be great to have that reviewed there by whomever reviews it here. It seems Obviously Correct™ since it deletes an unused file…

I have tested these changes on a converted system by building a new `gnome-initial-setup` `.deb` and installing it, then booting the system in the following ways:

* Add `gdm.force-initial-setup` to the kernel command line, then:
  * complete the FBE normally
  * enter test mode
  * enter demo mode, observing that the relevant settings changes still apply
* Add `gdm.force-initial-setup endless.live_boot` to the kernel command line, then follow the "Try Endless" flow.

There are still some warnings in the journal from `gnome-initial-setup-copy-worker` but I think they are spurious, and are already present in the normal FBE flow. eg. copying `.config/gnome-initial-setup-done` fails with `EEXIST` which I believe is caused by the `--existing-user` mode launched from all user sessions touching this stamp file too – it's a race. I avoided going down this unrelated rabbit hole as part of this change!

https://phabricator.endlessm.com/T16130


gnome-initial-setup/pages/live-chooser/eos-setup-live-user
```diff
@@ -178,6 +165,18 @@ class SetupLiveUser(object):
         self.update(GS_SCHEMA, ALLOW_UPDATES, GLib.Variant('b', False))
 
 
+def install_desktop_file():
+    """eos-installer is shipped in all images, but its desktop file is
+    (deliberately) only present in the GDM greeter's environment. For
+    image-booted live sessions, we symlink it into place."""
```
smspillaz: nitpick: PEP257 would require a short description (ending with a period) on the first line, followed by a longer description, with the closing triple-parens on the following line.

gnome-initial-setup/pages/live-chooser/eos-setup-live-user
```diff
@@ -178,6 +165,18 @@ class SetupLiveUser(object):
...
+    if not os.path.exists(LOCAL_DESKTOP_PATH):
+        log.info('ensuring %s exists', LOCAL_APPS_DIR)
+        os.makedirs(LOCAL_APPS_DIR, exist_ok=True)
+
+        log.info('symlinking %s to %s', EOS_INSTALLER, LOCAL_DESKTOP_PATH)
+        os.symlink(EOS_INSTALLER_PATH, LOCAL_DESKTOP_PATH)
```
smspillaz: I think a more pythonic way of doing this would be:

    os.makedirs(...)
    try:
        os.symlink(EOS_INSTALLER_PATH, LOCAL_DESKTOP_PATH)
    except OSError as error:
        if error.errno != errno.EEXIST:
            raise error

`os.path.exists` is technically racy, though I appreciate that this is a somewhat naff point.

gnome-initial-setup/pages/live-chooser/eos-setup-live-user
```diff
@@ -178,6 +165,18 @@ class SetupLiveUser(object):
         self.update(GS_SCHEMA, ALLOW_UPDATES, GLib.Variant('b', False))
 
 
+def install_desktop_file():
```
smspillaz: (By the way, thanks for making this a non-member function of the class - it was one of the things that was initially annoying me about this code when I first worked with it ;-) )

gnome-initial-setup/pages/live-chooser/eos-setup-live-user
```diff
@@ -49,48 +49,14 @@ GS_SCHEMA = 'org.gnome.software'
...
-
-    def format_keyfile(self):
-        keyfile_lines = []
-        for updates in self.updates.values():
-            keyfile_lines.append(updates.to_keyfile_str())
-
-        return '\n'.join(keyfile_lines)
+        value = variant.print_(False)
+        log.info('Updating %s: %s to %s', schema, key, value)
+        self.keyfile.set_string(schema.replace('.', '/'), key, value)
```
smspillaz: Calling `g_key_file_set_string` here with the stringified variant value seems a little dangerous. Is the value from `g_variant_print` always guaranteed to be the representation expected by `GKeyFile`?

gnome-initial-setup/pages/live-chooser/eos-setup-live-user
```diff
@@ -130,8 +128,12 @@ def install_desktop_file():
         os.symlink(EOS_INSTALLER_PATH, LOCAL_DESKTOP_PATH)
 
 
-def dq(s):
-    return '"{}"'.format(s)
+def replace_chrome_with_chromium(apps):
+    if GOOGLE_CHROME in apps:
```
smspillaz: nit: PEP257 would require a docstring here

gnome-initial-setup/pages/live-chooser/eos-setup-live-user
```diff
@@ -142,15 +144,14 @@ def remove_chrome_from_icon_grids():
...
             with open(path, 'r') as f:
-                grid = f.read()
+                grid = json.load(fp=f)
 
-            if GOOGLE_CHROME in grid:
+            if replace_chrome_with_chromium(grid[DESKTOP_GRID_ID]):
                 log.info("Updating %s", path)
-                grid = grid.replace(dq(GOOGLE_CHROME), dq(CHROMIUM_BROWSER))
 
                 with open(path, 'w') as f:
```
smspillaz: It isn't a part of this commit, but it'd be nice to avoid overloading `f` like this

gnome-initial-setup/pages/live-chooser/eos-setup-live-user
wjt: It's not about what GKeyFile expects, it's about what `dconf(1)` expects. `man 1 dconf` describes the format in passing:

```
       dump
           Dump an entire subpath to stdout. The output is in a keyfile-like
           format, with values in GVariant syntax.
```

[Here's the code that reads them.](https://git.gnome.org/browse/dconf/tree/bin/dconf-update.vala#n119) The [corresponding `dconf dump` code](https://git.gnome.org/browse/dconf/tree/bin/dconf-dump.vala#n1) uses exactly this method to write them.

gnome-initial-setup/pages/live-chooser/eos-setup-live-user
wjt: I don't really agree: this is outside the scope of the first `with` block.

gnome-initial-setup/pages/live-chooser/eos-setup-live-user
smspillaz: Makes sense - thanks for clarifying

gnome-initial-setup/pages/live-chooser/eos-setup-live-user
smspillaz: The second `with` block on line 151 is nested within the first on line 144.

...Unless my eyes are playing tricks on me?

wjt: They are! The body of the first `with` is a single line. Then a new `if` block appears.

```python
with open(path, 'r') as f:
    grid = json.load(fp=f)

if replace_chrome_with_chromium(grid[DESKTOP_GRID_ID]):
    log.info("Updating %s", path)

    with open(path, 'w') as f:
        json.dump(grid, fp=f)
```

smspillaz: I see - GitHub must have conveniently left out that `if` statement by suppressing context. Consider my comment irrelevant then :-)

wjt:
I pushed a couple of commits addressing @smspillaz's points that I haven't replied to above.

To a question elsewhere:

> how is `eos-transient-setup` invoked by `gnome-initial-setup` when we enter the demo mode? Is it just done the same way through `gis_pkexec`, or is it done through spawning the systemd service?

It's [called by the `eos-demo-mode` script](https://github.com/endlessm/gnome-initial-setup/pull/226/files#diff-c938d2d7fe614ae57a847d4b795bc89fR60).

gnome-initial-setup/eos-transient-setup
```diff
@@ -162,12 +162,12 @@ def install_installer_desktop_file():
...
     """
-    if not os.path.exists(LOCAL_DESKTOP_PATH):
-        log.info('ensuring %s exists', LOCAL_APPS_DIR)
-        os.makedirs(LOCAL_APPS_DIR, exist_ok=True)
-
+    os.makedirs(LOCAL_APPS_DIR, exist_ok=True)
+    try:
         log.info('symlinking %s to %s', EOS_INSTALLER, LOCAL_DESKTOP_PATH)
         os.symlink(EOS_INSTALLER_PATH, LOCAL_DESKTOP_PATH)
+    except FileExistsError:
```
smspillaz: Also works ✅

smspillaz:
> It's called by the eos-demo-mode script.

Thanks for clarifying!

PR #228, opened 2018-03-01
EOS-Video-Demo: start video playing when window is realized

erikos:
https://phabricator.endlessm.com/T19087

### DIFF
--- a/gnome-initial-setup/demo/eos-demo-video.c
+++ b/gnome-initial-setup/demo/eos-demo-video.c
@@ -52,6 +52,18 @@ on_demo_video_event (GtkWidget *widget,
 }
 
 static gboolean
+on_demo_video_realize (GtkWidget *widget,
+                       gpointer  user_data)
+{
+  GisDemoVideo *demo_video = GIS_DEMO_VIDEO (user_data);
+  GisDemoVideoPrivate *priv = gis_demo_video_get_instance_private (demo_video);
+
+  gst_element_set_state (priv->player, GST_STATE_PLAYING);
+
+  return FALSE;
+}
+
+static gboolean
 on_video_bus_event (GstBus     *bus,
                     GstMessage *msg,
                     gpointer   user_data)
@@ -106,8 +118,13 @@ create_demo_window (GisDemoVideo *demo_video)
                            demo_video,
                            G_CONNECT_AFTER);
 
+  g_signal_connect_object (GTK_WIDGET (priv->main_window),
+                           "realize",
+                           G_CALLBACK (on_demo_video_realize),
+                           demo_video,
+                           G_CONNECT_AFTER);
+
   gtk_widget_show_all (GTK_WIDGET (priv->main_window));
-  gst_element_set_state (priv->player, GST_STATE_PLAYING);
 
   gst_bus_add_watch (priv->bus, on_video_bus_event, demo_video);
 }


smspillaz:
LGTM

erikos:
Thanks Sam, to me too, though, it does not bring any improvements sadly. I tested with installing it and forcing the demo mode.

gnome-initial-setup/demo/eos-demo-video.c
```diff
@@ -106,8 +118,13 @@ create_demo_window (GisDemoVideo *demo_video)
                            demo_video,
                            G_CONNECT_AFTER);
 
+  g_signal_connect_object (GTK_WIDGET (priv->main_window),
```
rshuler: Is "GTK_WIDGET ()" needed here?  If so, shouldn't it also be used on the previous call to g_signal_connect_object?

smspillaz: Not strictly speaking no, though it never hurts (considering that the "realize" signal is on the widget itself).

That said, the follow-up commit connects to a signal only defined on GtkWindow, so this isn't relevant anymore and we can drop the cast.

gnome-initial-setup/demo/eos-demo-video.c
```diff
@@ -51,6 +51,21 @@ on_demo_video_event (GtkWidget *widget,
...
+                             gpointer    user_data)
+{
+  GtkWindow    *window = GTK_WINDOW (object);
+  GisDemoVideo *demo_video = GIS_DEMO_VIDEO (user_data);
+  GisDemoVideoPrivate *priv = gis_demo_video_get_instance_private (demo_video);
+
+  if (gtk_window_is_active (window))
+    gst_element_set_state (priv->player, GST_STATE_PLAYING);
+
+  return FALSE;
```
cosimoc: Pretty sure that property notification signals return `void`

gnome-initial-setup/demo/eos-demo-video.c
```diff
@@ -106,8 +121,13 @@ create_demo_window (GisDemoVideo *demo_video)
                            demo_video,
                            G_CONNECT_AFTER);
 
+  g_signal_connect_object (priv->main_window,
```
cosimoc: Will this not make it so that if someone e.g. alt-tabs back and forth the pipeline state will be set again? Not sure if it's a problem in practice.

gnome-initial-setup/demo/eos-demo-video.c
smspillaz: > still surprised that gcc does not warn about this ....

Ah, that's actually because the return type of the function was already `gboolean`

smspillaz: Fixed in any case

gnome-initial-setup/demo/eos-demo-video.c
cosimoc: Yeah, I think that'd be a good idea.

rshuler: In my testing, this change of signal does not help.

FWIW, as soon as I hit the Alt button, the video window is closed.

erikos: Yeah, the video does still not start at the beginning with this change. Starts at a similar time then before which suggests that the delay is introduced by another factor.

PR #229, opened 2018-03-02
Improve transition between FBE and eos-installer on eosinstaller images

wjt:
Previously, on reformatter-only images, the language page would advance to the live-chooser page, which would immediately hide the FBE and launch eos-installer. When that process exited, the live-chooser page would re-show the FBE and move to the language page.

As well as being awkward, if animations are enabled (as they are on most non-VM systems) then the transition back to the language page is not instantaneous, so you see a flash of the live-chooser page when eos-installer exits.

To solve this, move the hide-launch-wait-reshow logic to gis-page-util, and launch eos-installer directly from the language page when the user clicks Next. To avoid actually advancing to the next page, we (ab)use the ->apply() vfunc, which is intended to be used by pages like the keyboard page which need to asynchronously change system settings when you hit Next, delaying the transition to the next page, and blocking it in the event that the operation fails. In this case, the asynchronous operation is to launch eos-installer, wait for it to terminate, and always report that ->apply() failed.

This also incorporates a small bugfix that I've sent upstream on a [miscellaneous cleanup](https://bugzilla.gnome.org/show_bug.cgi?id=793833) ticket.

https://phabricator.endlessm.com/T14616

### DIFF
--- a/gnome-initial-setup/gis-page-util.c
+++ b/gnome-initial-setup/gis-page-util.c
@@ -1,6 +1,6 @@
 /* -*- Mode: C; tab-width: 8; indent-tabs-mode: nil; c-basic-offset: 8 -*-
  *
- * Copyright (C) 2017 Endless Mobile, Inc.
+ * Copyright © 2017–2018 Endless Mobile, Inc.
  *
  * This program is free software; you can redistribute it and/or
  * modify it under the terms of the GNU General Public License
@@ -13,12 +13,11 @@
  * GNU General Public License for more details.
  *
  * You should have received a copy of the GNU General Public License
- * along with this program; if not, write to the Free Software
- * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
- * 02110-1301, USA
+ * along with this program; if not, see <http://www.gnu.org/licenses/>.
  *
  * Written by:
  *     Mario Sanchez Prada <mario@endlessm.com>
+ *     Will Thompson <wjt@endlessm.com>
  */
 
 #include "config.h"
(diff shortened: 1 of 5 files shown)


gnome-initial-setup/gis-page-util.c
```diff
@@ -492,3 +491,94 @@ gis_page_util_show_demo_dialog (GisPage *page)
...
+      GtkWindow *toplevel = GTK_WINDOW (gtk_widget_get_toplevel (GTK_WIDGET (page)));
+      GtkWidget *message_dialog;
+
+      g_critical ("Error running the reformatter: %s", error->message);
+
+      message_dialog = gtk_message_dialog_new (toplevel,
+                                               GTK_DIALOG_MODAL | GTK_DIALOG_DESTROY_WITH_PARENT,
+                                               GTK_MESSAGE_ERROR,
+                                               GTK_BUTTONS_CLOSE,
+                                               _("Error running the reformatter: %s"), error->message);
```
joaquimrocha: Missing `TRANSLATORS` comment. Maybe not so important if this is just for Endless, but still probably worth it?

gnome-initial-setup/gis-page-util.c
```diff
@@ -492,3 +491,94 @@ gis_page_util_show_demo_dialog (GisPage *page)
...
+  gtk_dialog_run (dialog);
+  return G_SOURCE_REMOVE;
+}
+
+static void
+on_reformatter_exited (GTask  *task,
+                       GError *error)
+{
+  GisPage *page = GIS_PAGE (g_task_get_source_object (task));
+  GisDriver *driver = page->driver;
```
joaquimrocha: You don't really need to declare a variable for something that's only used once and in the next line.

gnome-initial-setup/gis-page-util.c
```diff
@@ -492,3 +491,94 @@ gis_page_util_show_demo_dialog (GisPage *page)
...
+
+  gis_driver_show_window (driver);
+
+  if (error == NULL)
+    {
+      g_task_return_boolean (task, TRUE);
```
joaquimrocha: I'd just early return here. Makes one less block: readibility++.

gnome-initial-setup/gis-page-util.c
```diff
@@ -492,3 +491,94 @@ gis_page_util_show_demo_dialog (GisPage *page)
...
+  const gchar *command = "/usr/lib/eos-installer/gnome-image-installer";
+  g_autoptr(GError) error = NULL;
+
+  launcher = g_subprocess_launcher_new (G_SUBPROCESS_FLAGS_NONE);
+  g_subprocess_launcher_setenv (launcher, "LANG", locale, TRUE);
+  subprocess = g_subprocess_launcher_spawn (launcher, &error, command, NULL);
+
+  if (error)
+    {
+      on_reformatter_exited (task, g_steal_pointer (&error));
```
joaquimrocha: Again you could early return here.

gnome-initial-setup/gis-page-util.c
```diff
@@ -492,3 +491,94 @@ gis_page_util_show_demo_dialog (GisPage *page)
...
+ *
+ * There is no corresponding _finish() function because (at present) neither
+ * caller actually cares about the result.
+ */
+void
+gis_page_util_run_reformatter (GisPage            *page,
+                               GAsyncReadyCallback callback,
+                               gpointer            user_data)
+{
+  g_autoptr(GTask) task = g_task_new (page, NULL, callback, user_data);
```
joaquimrocha: I don't understand why a `GTask` is needed here. AFAICS you are only using it to pass some info around, and never actually run any task with it, or did I miss something?

joaquimrocha: Also, if you didn't need the `GTask`, the `on_reformatter_exited` code could be simplified and its use made more consistent: sometimes it's verifying the error before calling it, others just calling it directly.

gnome-initial-setup/gis-page-util.c
wjt: ok.

gnome-initial-setup/gis-page-util.c
wjt: ok.

gnome-initial-setup/gis-page-util.c
wjt: ok.

gnome-initial-setup/gis-page-util.c
wjt: I'm not sure this is easier to read with an early return, but ok.

gnome-initial-setup/gis-page-util.c
wjt: I am using the GTask to:

a) represent an ongoing async operation
b) store a reference to the page, a callback, and some user data
c) call that callback when the operation is complete

It's true that there's no _finish() function but I don't think writing my own struct to hold the callback and call it would be a net win.

wjt: I don't think the `on_reformatter_exited` code can be made simpler because it is called in two cases:

1. spawning the task failed entirely
2. having been successfully spawned, the task exited some time later, and we check its exit status

joaquimrocha: I don't know. It feels a bit weird to me that you're essentially using a `GTask` as a struct for storing a page and a callback. So IMO a struct with those two members would just be simpler (reading the code I kept expecting to see something running the `GTask`).

But I won't block this if you feel confident that this is a good idea...

gnome-initial-setup/gis-page-util.c
```diff
@@ -505,31 +505,34 @@ on_reformatter_exited (GTask  *task,
...
-                                               GTK_MESSAGE_ERROR,
-                                               GTK_BUTTONS_CLOSE,
-                                               _("Error running the reformatter: %s"), error->message);
-      g_idle_add_full (G_PRIORITY_DEFAULT_IDLE, run_dialog_cb,
-                       message_dialog, (GDestroyNotify) gtk_widget_destroy);
-
-      g_task_return_error (task, g_steal_pointer (&error));
-    }
+  GtkWindow *toplevel = GTK_WINDOW (gtk_widget_get_toplevel (GTK_WIDGET (page)));
+  GtkWidget *message_dialog;
```
joaquimrocha: BTW, reluctantly I usually move the definitions to the top as is common in glib, but if that's not enforced in g-i-s, I am okay with it.

wjt: I would rather have the `else` block (to introduce a new scope), or use mixed definitions, rather than move the declarations to the top of function.

PR #230, opened 2018-04-24
Reimplement the Display page using org.gnome.Mutter.DisplayConfig

mariospr:
https://phabricator.endlessm.com/T21591

### DIFF
--- a/gnome-initial-setup/pages/display/gis-display-page.c
+++ b/gnome-initial-setup/pages/display/gis-display-page.c
@@ -30,21 +30,22 @@
 
 #include "display-resources.h"
 
+#include "cc-display-config-manager-dbus.h"
+#include "cc-display-config.h"
+
 #include <glib/gi18n.h>
 #include <gio/gio.h>
-#include <libgnome-desktop/gnome-rr.h>
-#include <libgnome-desktop/gnome-rr-config.h>
 
 typedef struct {
   GtkWidget *overscan_on;
   GtkWidget *overscan_off;
   GtkWidget *overscan_default_selection;
 
-  GnomeRRScreen *screen;
-  GnomeRRConfig *current_config;
-  GnomeRROutputInfo *current_output;
+  CcDisplayConfigManager *manager;
+  CcDisplayConfig *current_config;
+  CcDisplayMonitor *current_output;
 
-  guint screen_changed_id;
+  gulong screen_changed_id;
 } GisDisplayPagePrivate;
 
 G_DEFINE_TYPE_WITH_PRIVATE (GisDisplayPage, gis_display_page, GIS_TYPE_PAGE);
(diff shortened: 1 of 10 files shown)


gnome-initial-setup/pages/display/gis-display-page.c
```diff
@@ -116,42 +130,22 @@ toggle_overscan (GisDisplayPage *page,
...
-  char *output_name;
 
-  output_name = gnome_rr_output_info_get_name (priv->current_output);
-  output = gnome_rr_screen_get_output_by_name (priv->screen, output_name);
-
-  if (!output)
-    return FALSE;
-
-  return gnome_rr_output_supports_underscanning (output);
+  read_screen_config (page);
```
cosimoc: Why do you call this explicitly? I would expect we get a changed signal after we set the new configuration, and we'd update it that way.

gnome-initial-setup/pages/display/gis-display-page.c
```diff
@@ -72,41 +70,57 @@ set_toggle_options_from_config (GisDisplayPage *page)
...
+  priv->current_output = NULL;
 
-  current = gnome_rr_config_new_current (priv->screen, NULL);
-  gnome_rr_config_ensure_primary (current);
-  priv->current_config = current;
+  current = cc_display_config_manager_get_current (priv->manager);
+  if (!current)
+    {
+      g_critical ("Could not get primary output information. Hiding overscan page.");
+      return FALSE;
```
cosimoc: I don't think we'd hide it from this code path -- the call to `gtk_widget_set_visible()` is at the end of this function.

gnome-initial-setup/pages/display/gis-display-page.c
```diff
@@ -72,41 +70,57 @@ set_toggle_options_from_config (GisDisplayPage *page)
...
-  int i;
 
-  gnome_rr_screen_refresh (priv->screen, NULL);
+  if (!priv->current_output)
+    return FALSE;
+
+  return cc_display_monitor_supports_underscanning (priv->current_output);
+}
+
+static gboolean
```
cosimoc: It does not look like you make any use of this return value.

gnome-initial-setup/pages/display/gis-display-page.c
mariospr: Oops, right! This was an overlook while refactoring. I'll change it to make sure we `goto` the end in this case and set visible accordingly.

mariospr: ...or better: move the code that sets the visibility out of this function and into where it used to be before: `gis_display_page_constructed`

mariospr: Actually, I can't move it exactly there because of the asynchronous initialization of the CcDisplayManager, so scrap that.

gnome-initial-setup/pages/display/gis-display-page.c
mariospr: I will use it in the updated PR

gnome-initial-setup/pages/display/gis-display-page.c
mariospr: I'm simply preserving the behaviour from the previous implementation, see https://github.com/endlessm/gnome-initial-setup/pull/230/commits/dad0e210cb7ebc4bda80e3586eac19460041f26b#diff-1591211bb91441665f14a68847d53cf9L95

But I agree it might not be needed, I'll try removing it and see how it works

mariospr: It was not needed indeed, removed

mariospr:
Updated the PR. Please **do NOT merge** if the review is approved. I'll squash those fixups and merge myself, thanks.

gnome-initial-setup/pages/display/gis-display-page.c
```diff
@@ -116,44 +119,22 @@ toggle_overscan (GisDisplayPage *page,
                  gboolean value)
 {
   GisDisplayPagePrivate *priv = gis_display_page_get_instance_private (page);
-  GError *error;
+  gboolean current_value = FALSE;
```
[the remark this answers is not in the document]
mariospr: Ooops! Fixed now.

PR #231, opened 2018-04-25
Don't use a previously introduced password for non protected accounts

mariospr:
https://phabricator.endlessm.com/T22042

### DIFF
--- a/gnome-initial-setup/pages/password/gis-password-page.c
+++ b/gnome-initial-setup/pages/password/gis-password-page.c
@@ -206,6 +206,26 @@ validate (GisPasswordPage *page)
   return FALSE;
 }
 
+static void
+page_visibility_changed (GtkWidget  *w,
+                         GParamSpec *pspec,
+                         gpointer    unused)
+{
+  /* Make sure to reset the password page in this case, to prevent situations
+   * where a previously introduced password, that is no longer required, would
+   * end up being used anyway when creating the user's account.
+   */
+  if (!gtk_widget_get_visible (w)) {
+    GisPasswordPage *page = GIS_PASSWORD_PAGE (w);
+    GisPasswordPagePrivate *priv = gis_password_page_get_instance_private (page);
+
+    gtk_entry_set_text (GTK_ENTRY (priv->password_entry), "");
+    gtk_entry_set_text (GTK_ENTRY (priv->confirm_entry), "");
+    gtk_entry_set_text (GTK_ENTRY (priv->reminder_entry), "");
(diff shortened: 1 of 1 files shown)


gnome-initial-setup/gis-page.c
```diff
@@ -445,3 +445,10 @@ gis_page_shown (GisPage *page)
   if (GIS_PAGE_GET_CLASS (page)->shown)
     GIS_PAGE_GET_CLASS (page)->shown (page);
 }
+
+void
+gis_page_reset (GisPage *page)
```
cosimoc: It does not feel like this generic method is actually needed here. You could simply add a method to the password page and call it from the account page directly, but I wonder if you could not simply do the cleanup from a visibility signal handler in the password page. The account page already calls `set_password_page_visibility()`...

mariospr: The problem is that all that `set_password_page_visibility()` does is to call `gtk_widget_set_visible()` over the password page, which is fine because it's a `GtkWidget` after all, but the account page can't access any of the internal data that needs to be reset in there, that is, password entry, password confirmation entry, reminder text and the "show password" checkbox.

Those operations deal with internal properties of the password page that need to be dealt with internally (unless we break that encapsulation, but I don't think that's worth it.)

mariospr:
Updated the PR after clarification from @cosimoc on Slack

gnome-initial-setup/pages/password/gis-password-page.c
```diff
@@ -206,6 +206,26 @@ validate (GisPasswordPage *page)
   return FALSE;
 }
 
+static void
+page_visibility_changed (GtkWidget       *w,
+                         GParamSpec      *pspec,
+                         GisPasswordPage *page)
```
cosimoc: You’re not actually passing the page as user_data to the signal callback, so this is technically not correct.
Also you declare a variable with the same name in th inner block.

gnome-initial-setup/pages/password/gis-password-page.c
```diff
@@ -206,6 +206,26 @@ validate (GisPasswordPage *page)
   return FALSE;
 }
 
+static void
+page_visibility_changed (GtkWidget  *w,
+                         GParamSpec *pspec,
+                         gpointer   *unused)
```
cosimoc: Sorry for the nitpicking, but this should be `gpointer` (`gpointer *` makes it a double pointer)

mariospr: WTF... this is what happens when I have my brain debugging things on our insane apps grid I guess...

PR #232, opened 2018-04-26
password: Relax requirements to validate the password page

mariospr:
Upstream won't let users create an account if the password is considered
too weak (i.e. "bad"), but we don't want to be that restrictive, so let's
remove that requirement and simply consider a password good enough as
long as it's a non-empty string.

https://phabricator.endlessm.com/T22042

### DIFF
--- a/gnome-initial-setup/pages/password/gis-password-page.c
+++ b/gnome-initial-setup/pages/password/gis-password-page.c
@@ -51,7 +51,6 @@ struct _GisPasswordPagePrivate
   GDBusProxy *input_source_manager;
 
   gboolean valid_confirm;
-  gboolean valid_password;
   guint timeout_id;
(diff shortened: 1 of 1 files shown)


gnome-initial-setup/pages/password/gis-password-page.c
```diff
@@ -187,8 +186,7 @@ validate (GisPasswordPage *page)
   gtk_label_set_label (GTK_LABEL (priv->confirm_explanation), "");
   priv->valid_confirm = FALSE;
 
-  priv->valid_password = (strength_level > 1);
```
cosimoc: (I merged this patch to avoid a lot of back-and-forth since it looks good, but it would have been simpler to just change this assignment to `TRUE` and keep the patch small.)

PR #237, opened 2018-07-18
T23186

joaquimrocha:
https://phabricator.endlessm.com/T23327

### DIFF
--- a/gnome-initial-setup/gis-driver.c
+++ b/gnome-initial-setup/gis-driver.c
@@ -90,6 +90,8 @@ struct _GisDriverPrivate {
   gboolean small_screen;
   gboolean hidden;
 
+  GKeyFile *vendor_conf_file;
+
   /* Cancelled on shutdown */
   GCancellable *cancellable;
 };
@@ -212,6 +214,7 @@ gis_driver_finalize (GObject *object)
 
   g_clear_object (&priv->user_account);
   g_clear_object (&priv->cancellable);
+  g_clear_pointer (&priv->vendor_conf_file, g_key_file_free);
 
   G_OBJECT_CLASS (gis_driver_parent_class)->finalize (object);
 }
@@ -371,6 +374,31 @@ void
 gis_driver_locale_changed (GisDriver *driver)
 {
   g_signal_emit (G_OBJECT (driver), signals[LOCALE_CHANGED], 0);
+}
+
+static void
+load_vendor_conf_file (GisDriver *driver)
+{
+  GisDriverPrivate *priv = gis_driver_get_instance_private (driver);
+  g_autoptr(GError) error = NULL;
+  g_autoptr(GKeyFile) vendor_conf_file = g_key_file_new ();
+
+  if(!g_key_file_load_from_file (vendor_conf_file, VENDOR_CONF_FILE,
+                                 G_KEY_FILE_NONE, &error))
+    {
+      if (!g_error_matches (error, G_FILE_ERROR, G_FILE_ERROR_NOENT))
+        g_warning ("Could not read file %s: %s", VENDOR_CONF_FILE, error->message);
+      return;
+    }
+
+  priv->vendor_conf_file = g_steal_pointer (&vendor_conf_file);
+}
+
+GKeyFile *
+gis_driver_get_vendor_conf_file (GisDriver *driver)
+{
+  GisDriverPrivate *priv = gis_driver_get_instance_private (driver);
+  return priv->vendor_conf_file;
 }
 
 GisDriverMode
@@ -809,6 +837,8 @@ gis_driver_init (GisDriver *driver)
   screen = gdk_screen_get_default ();
 
   priv->small_screen = screen_is_small (screen);
+
+  load_vendor_conf_file (driver);
 
   g_signal_connect (screen, "size-changed",
                     G_CALLBACK (screen_size_changed), driver);
--- a/gnome-initial-setup/gnome-initial-setup.c
+++ b/gnome-initial-setup/gnome-initial-setup.c
@@ -156,6 +156,91 @@ destroy_pages_after (GisAssistant *assistant,
   }
 }
 
+static gint
+compare_pages_order (gconstpointer a,
+                     gconstpointer b,
+                     gpointer data)
+{
+  GHashTable *sort_table = data;
+  const gchar *id_a = ((PageData *) a)->page_id;
+  const gchar *id_b = ((PageData *) b)->page_id;
+  gint order_a = GPOINTER_TO_INT (g_hash_table_lookup (sort_table, id_a));
+  gint order_b = GPOINTER_TO_INT (g_hash_table_lookup (sort_table, id_b));
+
+  return order_a - order_b;
+}
+
+static void
+reorder_pages (GisDriver *driver)
+{
+  gint num_pages;
+  GKeyFile *key_file = gis_driver_get_vendor_conf_file (driver);
+  gsize num_ordered_pages = 0;
+  g_auto(GStrv) conf_pages_order = NULL;
+  g_autoptr(GHashTable) sort_table = NULL; /* (owned) (element-type utf8 uint) */
+  g_autoptr(GError) error = NULL;
+
+  if (key_file == NULL)
+    return;
+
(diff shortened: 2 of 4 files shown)


gnome-initial-setup/gnome-initial-setup.c
```diff
@@ -156,6 +158,81 @@ destroy_pages_after (GisAssistant *assistant,
...
+
+  for (num_pages = 0; page_table[num_pages].page_id != NULL; ++num_pages) {
+    g_hash_table_insert (sort_table, (gchar *) page_table[num_pages].page_id,
+                         GUINT_TO_POINTER (PAGES_DEFAULT_SORT_WEIGHT));
+  }
+
+  /* assign the new sort weights from the conf file */
+  for (guint i = 0; conf_pages_order[i] != NULL; ++i) {
+    gchar *page_id = conf_pages_order[i];
+    if (g_hash_table_lookup (sort_table, page_id) == NULL) {
```
wjt: `g_hash_table_contains (sort_table, page_id)`? It's equivalent because the default sort weight is non-zero but I think it's clearer.

gnome-initial-setup/gnome-initial-setup.c
```diff
@@ -156,6 +158,81 @@ destroy_pages_after (GisAssistant *assistant,
...
+      g_debug ("No page with id '%s' found while sorting the pages!", page_id);
```
wjt: Are `g_debug` messages shown in the journal by default? If not I think this deserves to be an info/warning since it represents a misconfiguration.

gnome-initial-setup/gnome-initial-setup.c
```diff
@@ -156,6 +158,81 @@ destroy_pages_after (GisAssistant *assistant,
...
+      continue;
+    }
+
+    g_hash_table_insert (sort_table, page_id, GUINT_TO_POINTER (i));
```
wjt: What if there are more than 100 pages in the config file? I suggest that rather than define PAGES_DEFAULT_SORT_WEIGHT, you pass a `length` parameter to `g_key_file_get_string_list` and then use that as the default sort weight – it's guaranteed to be 1 higher than the index of any page in the config file.

gnome-initial-setup/gnome-initial-setup.c
```diff
@@ -156,6 +158,81 @@ destroy_pages_after (GisAssistant *assistant,
...
 
+static gint
+compare_pages_order (gconstpointer a,
+                     gconstpointer b,
+                     gpointer data)
+{
+  GHashTable *sort_table = data;
+  const gchar *id_a = ((PageData *) a)->page_id;
+  const gchar *id_b = ((PageData *) b)->page_id;
+  gint order_a = (gint) GPOINTER_TO_UINT (g_hash_table_lookup (sort_table, id_a));
```
wjt: Why use `GPOINTER_TO_UINT` then cast to `gint`? Just use `GPOINTER_TO_INT` (and `GINT_TO_POINTER` when building the hash table).

gnome-initial-setup/gnome-initial-setup.c
wjt: (If you follow this suggestion you will have to clamp the `gsize length` at `G_MAXINT` since gsize may be larger than gint.)

gnome-initial-setup/gnome-initial-setup.c
```diff
@@ -156,6 +158,81 @@ destroy_pages_after (GisAssistant *assistant,
...
+    gchar *page_id = conf_pages_order[i];
+    if (g_hash_table_lookup (sort_table, page_id) == NULL) {
+      g_debug ("No page with id '%s' found while sorting the pages!", page_id);
+      continue;
+    }
+
+    g_hash_table_insert (sort_table, page_id, GUINT_TO_POINTER (i));
+  }
+
+  g_qsort_with_data (page_table,
```
wjt: I had to check the documentation to see if this is a stable sort (ie, preserves the existing order of pages which compare equal according to `compare_pages_order`). It is, since GLib 2.32, but please add a comment to this effect.

joaquimrocha: I know it's stable and I mentioned the order being kept in the commit message. Also, g-i-s requires glib 2.53.0.
In any case, I don't think a comment makes a lot of difference here. If we need to make sure this doesn't ship if suddenly g-i-s is changed to use glib < 2.32 (very unlikely), we should use some `ifdefs` or whatever with the glib version.

wjt: I'm not worried that it will break; and of course not worried that we will suddenly downgrade to a prehistoric GLib. I was worried because I know that Quicksort is, in general, not stable. `man 3 qsort` says:

> If two members compare as equal, their order in the sorted array is undefined.

So upon reading the code, it was not clear to me that the claim in the commit message:

> […] rest of the pages, in the default order they had among themselves

was true. This is why I suggested a comment. If you think it is obvious that `g_qsort_with_data` preserves order for equal elements, okay.

gnome-initial-setup/gis-driver.c
```diff
@@ -373,6 +376,31 @@ gis_driver_locale_changed (GisDriver *driver)
...
+{
+  GisDriverPrivate *priv = gis_driver_get_instance_private (driver);
+  g_autoptr(GError) error = NULL;
+  g_autoptr(GKeyFile) vendor_conf_file = g_key_file_new ();
+
+  if(!g_key_file_load_from_file (vendor_conf_file, VENDOR_CONF_FILE,
+                                 G_KEY_FILE_NONE, &error))
+    {
+      if (!g_error_matches (error, G_IO_ERROR, G_IO_ERROR_NOT_FOUND))
+        g_debug ("Could not read file %s: %s", VENDOR_CONF_FILE, error->message);
```
wjt: I just noticed on a second reading that the old code this replaces was:

```c
    if (!g_error_matches (error, G_FILE_ERROR, G_FILE_ERROR_NOENT))
      g_warning ("Could not read file %s: %s", VENDOR_CONF_FILE, error->message);
```

a. was the old GError domain/code wrong?
b. should this not still be a warning?

gnome-initial-setup/gnome-initial-setup.c
```diff
@@ -58,6 +58,8 @@
 #define VENDOR_PAGES_GROUP "pages"
 #define VENDOR_PAGES_SKIP_KEY "skip"
 
+#define PAGES_DEFAULT_SORT_WEIGHT 100
```
wjt: This `#define` is now unused and should be removed.

gnome-initial-setup/gis-driver.c
joaquimrocha: I don't know if it was wrong, but I've always used `G_IO_ERROR_NOT_FOUND` for such cases, and I have just double checked that it's working as expected.
I will update the debug message to a warning though.

gnome-initial-setup/gnome-initial-setup.c
joaquimrocha: Sorry about that.

gnome-initial-setup/gis-driver.c
wjt: The documentation for `g_key_file_load_from_file` says:

> This function will never return a G_KEY_FILE_ERROR_NOT_FOUND error. If the file is not found, G_FILE_ERROR_NOENT is returned.

And I have confirmed that the error returned is G_FILE_ERROR, G_FILE_ERROR_NOT_FOUND:

```python
In [1]: from gi.repository import GLib
In [4]: GLib.KeyFile().load_from_file('/does/not/exist', GLib.KeyFileFlags.NONE)
---------------------------------------------------------------------------
Error                                     Traceback (most recent call last)
<ipython-input-4-865cae36c2e6> in <module>()
----> 1 GLib.KeyFile().load_from_file('/does/not/exist', GLib.KeyFileFlags.NONE)

Error: g-file-error-quark: No such file or directory (4)
```

So the old code was correct and this is wrong

joaquimrocha: Shit. You're actually right and I was testing it the wrong way. Sorry for the noise and thanks for not letting this pass!

PR #239, opened 2018-07-23
account: Check configuration to hide password switch if needed

joaquimrocha:
If a vendor wants to have passwordless accounts by default, then they
should be able to set that in the vendor configuration file.
This patch allows that by using the "show-password-switch" option in the
"page.account" group.

https://phabricator.endlessm.com/T23326

### DIFF
--- a/gnome-initial-setup/pages/account/gis-account-page.c
+++ b/gnome-initial-setup/pages/account/gis-account-page.c
@@ -32,6 +32,9 @@
 #include <glib/gi18n.h>
 #include <gio/gio.h>
 
+#define CONFIG_ACCOUNT_GROUP "page.account"
+#define CONFIG_ACCOUNT_SHOW_PASSWORD_SWITCH_KEY "show-password-switch"
+
 struct _GisAccountPagePrivate
 {
   GtkWidget *page_local;
@@ -102,6 +105,36 @@ on_validation_changed (gpointer        page_area,
   update_page_validation (page);
 }
 
+
+static void
+hide_password_switch_if_needed (GisAccountPage *page)
+{
+  GKeyFile *conf_file = gis_driver_get_vendor_conf_file (GIS_PAGE (page)->driver);
+  GisAccountPagePrivate *priv = gis_account_page_get_instance_private (page);
+  g_autoptr(GError) error = NULL;
+  gboolean show_password_switch;
+
+  /* do nothing if the configuration file doesn't exist */
+  if (conf_file == NULL)
+    return;
+
+  /* check the conf file to see if the password switch should be shown/hidden */
+  show_password_switch = g_key_file_get_boolean (conf_file, CONFIG_ACCOUNT_GROUP,
+                                                 CONFIG_ACCOUNT_SHOW_PASSWORD_SWITCH_KEY,
+                                                 &error);
+  if (error == NULL) {
+    gis_account_page_local_show_password_switch (GIS_ACCOUNT_PAGE_LOCAL (priv->page_local),
+                                                 show_password_switch);
+    return;
+  }
+
+  if (!g_error_matches (error, G_KEY_FILE_ERROR, G_KEY_FILE_ERROR_KEY_NOT_FOUND) &&
+      !g_error_matches (error, G_KEY_FILE_ERROR, G_KEY_FILE_ERROR_GROUP_NOT_FOUND))
+    g_warning ("Couldn't get the value for key '%s' in group [%s]: %s",
+               CONFIG_ACCOUNT_SHOW_PASSWORD_SWITCH_KEY, CONFIG_ACCOUNT_GROUP,
+               error->message);
+}
+
 static void
 set_mode (GisAccountPage *page,
           UmAccountMode   mode)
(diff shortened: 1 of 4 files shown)


gnome-initial-setup/pages/account/gis-account-page.c
```diff
@@ -102,6 +102,32 @@ on_validation_changed (gpointer        page_area,
...
+  GKeyFile *conf_file = gis_driver_get_vendor_conf_file (GIS_PAGE (page)->driver);
+  GisAccountPagePrivate *priv = gis_account_page_get_instance_private (page);
+  g_autoptr(GError) error = NULL;
+
+  /* do nothing if the configuration file doesn't exist */
+  if (conf_file == NULL)
+    return;
+
+  /* check the conf file to see if the password switch should be hidden */
+  if (!g_key_file_get_boolean (conf_file, "page.account", "show-password-switch", &error) &&
```
cosimoc: It's a bit odd that the other use we have with this uses a different namespace (`Welcome` for the branding-welcome page if I'm not mistaken.)

joaquimrocha: I saw that but it doesn't invalidate this code, right? I actually think that _Welcome_ as a group that refers to the _branding-welcome_ can be misleading, and we should change it to be _page.branding-welcome_ instead. I can do that in a different PR.

wjt: It might be worth encapsulating the vendor config keyfile in a class so that all the groups and keys are maintained in one place. Right now it's a little hard to see what's being used where. I did this in the reformatter for the unattended configuration https://github.com/endlessm/eos-installer/blob/master/gnome-image-installer/util/gis-unattended-config.c and think it's all more legible as a result.

I think it's good to name the group after the page id – either `account` or `page.account` seems fine to me. It's unfortunate that the `branding-welcome` page uses `Welcome` but while we can make it check a new group/key to match a new convention, we still need to check the old group/key for backwards-compatibility with old vendor config files.

gnome-initial-setup/pages/account/gis-account-page.c
```diff
@@ -102,6 +102,32 @@ on_validation_changed (gpointer        page_area,
...
+      error == NULL) {
+    gis_account_page_local_show_password_switch (GIS_ACCOUNT_PAGE_LOCAL (priv->page_local),
+                                                 FALSE);
+    return;
+  }
+
+  if (!g_error_matches (error, G_KEY_FILE_ERROR, G_KEY_FILE_ERROR_KEY_NOT_FOUND) &&
+      !g_error_matches (error, G_KEY_FILE_ERROR, G_KEY_FILE_ERROR_GROUP_NOT_FOUND))
+    g_warning ("Couldn't get the value for key 'show-password-switch' in group "
```
wjt: Please make `show-password-switch` and `page.account` into `#define`s or `static const char *`s rather than open-coding them in this message and the lookup above.

gnome-initial-setup/pages/account/gis-account-page.c
wjt: Or, at least, put all the `#define`s for group & key names into one header, which can serve as the definition for what can be put in the keyfile. Right now it's smeared across the code.

joaquimrocha: @wjt I understand your suggestion, but what you did for the installer seems to suggest you had a more complex use-case than just retrieving configuration. I thought that not confining pages to using a "proxy" for the config here was a good thing, at least until we realize all the needs the pages have. I certainly don't want to do a wrapper for all the methods in `GKeyFile` until we understand if those are needed.

Since all pages are in the tree and no keys are being shared at the moment, I think it's okay to let pages worry about their keys (but I will move the group/keys into defines in the page's code here, if that makes it a bit more organized).

gnome-initial-setup/pages/account/gis-account-page.c
```diff
@@ -102,6 +105,34 @@ on_validation_changed (gpointer        page_area,
...
+  /* check the conf file to see if the password switch should be hidden */
+  if (!g_key_file_get_boolean (conf_file, CONFIG_ACCOUNT_GROUP,
+                               CONFIG_ACCOUNT_SHOW_PASSWORD_SWITCH_KEY, &error) &&
+      error == NULL) {
+    gis_account_page_local_show_password_switch (GIS_ACCOUNT_PAGE_LOCAL (priv->page_local),
+                                                 FALSE);
+    return;
+  }
+
+  if (!g_error_matches (error, G_KEY_FILE_ERROR, G_KEY_FILE_ERROR_KEY_NOT_FOUND) &&
```
wjt: On a second reading – the logic is wrong if you define `[page.account] show-password-switch = true`. In that case, `!g_key_file_get_boolean(...)` is FALSE; the block above is skipped; and we will crash either here or in the `g_warning` arguments because `error` is `NULL`. Please test all four cases: no key, `true`, `false`, `garbage`. There should be a warning only in the last case.

joaquimrocha: Yeah, apparently I didn't tested the `true` value with the final code. It won't crash in the `g_error_matches` since it returns `FALSE` if error is `NULL`, but it does crash on accessing it in the `g_warning` formatting. I have updated the PR.

PR #242, opened 2018-07-30
T23405 debian master on meson

joaquimrocha:
https://phabricator.endlessm.com/T23405


debian/control
```diff
@@ -40,6 +39,7 @@ Build-Depends: debhelper (>= 11),
                libattr1-dev,
                libevince-dev,
                libgnome-keyring-dev,
+               meson (>= 0.46),
```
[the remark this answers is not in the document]
joaquimrocha: True. Number confusion mistake. Thanks.

wjt:
Interestingly when built with `jenkins-obs` (I patched my local copy to disable the autotools build) this generates a Debian-native package:

```
dpkg-buildpackage: info: source-only upload: Debian-native package
...
Sending    gnome-initial-setup_3.28.0+dev184.457d45a5-2.dsc
Sending    gnome-initial-setup_3.28.0+dev184.457d45a5-2.tar.gz
```

rather than the orig tarball + debian/ tarball used before. Probably harmless in our case since that better reflects how the package is maintained anyway. I think this is probably common to all meson projects fed through jenkins-obs.

debian/control
```diff
@@ -40,6 +39,7 @@ Build-Depends: debhelper (>= 11),
                libattr1-dev,
                libevince-dev,
                libgnome-keyring-dev,
+               meson (>= 0.47),
```
wjt: You need to apply both these changes to `debian/control.in`. They are currently lost during the source package build process (which regenerates `debian/control` from `.in`) so the build fails on OBS: https://obs-master.endlessm-sf.com/package/live_build_log/home:wjt:branches:eos:master:endless/gnome-initial-setup/endless/x86_64

You'll also notice from that log that debhelper seems to rerun autoconf, even though it does eventually use the meson build. Another reason to drop the autotools build system from our fork?

joaquimrocha: >  debian/control.in

Weird I missed that. It's a bit misleading that both are in the source control.

> You'll also notice from that log that debhelper seems to rerun autoconf, even though it does eventually use the meson build. Another reason to drop the autotools build system from our fork?

I did notice that but wasn't sure if it was really running autotools or it was some Debian specific autotools stuff (I don't know enough about the Debian build system). Let me include the commit that nukes autotools.

PR #246, opened 2018-08-03
T23325

joaquimrocha:
https://phabricator.endlessm.com/T23325

### DIFF
--- a/gnome-initial-setup/pages/timezone/gis-timezone-page.c
+++ b/gnome-initial-setup/pages/timezone/gis-timezone-page.c
@@ -224,6 +224,10 @@ get_location_from_geoclue_async (GisTimezonePage *page)
 {
   GisTimezonePagePrivate *priv = gis_timezone_page_get_instance_private (page);
 
+  /* cancel any on-going requests, and set up a new cancellable */
+  stop_geolocation_if_needed (page);
+  g_set_object (&priv->geoclue_cancellable, g_cancellable_new ());
+
   gclue_simple_new (DESKTOP_ID,
                     GCLUE_ACCURACY_LEVEL_CITY,
                     priv->geoclue_cancellable,
@@ -463,8 +467,6 @@ gis_timezone_page_constructed (GObject *object)
   priv->clock_format = get_default_time_format ();
   g_settings_set_enum (settings, CLOCK_FORMAT_KEY, priv->clock_format);
   g_object_unref (settings);
-
-  priv->geoclue_cancellable = g_cancellable_new ();
 
   set_location (page, NULL);
 
--- a/gnome-initial-setup/pages/timezone/gis-timezone-page.c
+++ b/gnome-initial-setup/pages/timezone/gis-timezone-page.c
@@ -74,12 +74,15 @@ struct _GisTimezonePagePrivate
   gboolean in_search;
 
   gulong map_location_changed_id;
+  gulong network_monitor_handler_id;
 };
 typedef struct _GisTimezonePagePrivate GisTimezonePagePrivate;
 
 G_DEFINE_TYPE_WITH_PRIVATE (GisTimezonePage, gis_timezone_page, GIS_TYPE_PAGE);
 
 G_DEFINE_AUTO_CLEANUP_FREE_FUNC(locale_t, freelocale, NULL)
+
+static void stop_geolocation_if_needed (GisTimezonePage *page);
 
 static GDesktopClockFormat
 get_default_time_format (void)
@@ -149,6 +152,8 @@ set_location (GisTimezonePage  *page,
   GisTimezonePagePrivate *priv = gis_timezone_page_get_instance_private (page);
 
   g_clear_pointer (&priv->current_location, gweather_location_unref);
+
+  stop_geolocation_if_needed (page);
 
   gtk_widget_set_visible (priv->search_overlay, (location == NULL));
   gis_page_set_complete (GIS_PAGE (page), (location != NULL));
@@ -385,7 +390,7 @@ entry_mapped (GtkWidget *widget,
 }
 
 static void
-stop_geolocation (GisTimezonePage *page)
+stop_geolocation_if_needed (GisTimezonePage *page)
 {
   GisTimezonePagePrivate *priv = gis_timezone_page_get_instance_private (page);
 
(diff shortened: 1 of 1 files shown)


gnome-initial-setup/pages/timezone/gis-timezone-page.c
```diff
@@ -445,6 +477,13 @@ gis_timezone_page_constructed (GObject *object)
   priv->map_location_changed_id =
     g_signal_connect (priv->map, "location-changed",
                       G_CALLBACK (map_location_changed), page);
+  g_signal_connect (priv->network_monitor, "network-changed",
```
wjt: You'll want to disconnect this signal somewhere other than in its callback – what if the page is disposed, then the callback is called?

gnome-initial-setup/pages/timezone/gis-timezone-page.c
```diff
@@ -469,15 +508,6 @@ gis_timezone_page_locale_changed (GisPage *page)
   gis_page_set_title (GIS_PAGE (page), _("Time Zone"));
 }
 
-static void
-gis_timezone_page_shown (GisPage *page)
```
wjt: I assume that the point of this was to not modify the timezone while the user is trying to set it manually. What should happen in the following sequence of events?

* Network connection comes up
* This page asks Geoclue for location
* User advances to this page, and manually chooses location A
* Geoclue replies with location B

Should we overwrite the user's choice? I think that is what will happen with this patch.

gnome-initial-setup/pages/timezone/gis-timezone-page.c
joaquimrocha: I did forget to clear the `network_monitor` (I've just added that). If I do clear it on dispose, then it disconnects its signals, meaning there'll be no chance that the page is disposed and the callback is called, right?

gnome-initial-setup/pages/timezone/gis-timezone-page.c
joaquimrocha: Yeah, as explained in the commit message, this cannot happen. The implementation cancels any ongoing geoclue requests if a location is set (and a location can be set by the user, so, if they are quick enough, it'll be the users' location to be used and they will never get it reassigned).

gnome-initial-setup/pages/timezone/gis-timezone-page.c
wjt: What if something else holds a ref to the same, shared `network_monitor`? (Why not disconnect the signal handler before clearing it?)

wjt:
A higher-level comment: looking more closely at the geoclue API, it seems that it *should* be possible to set up the client before the network becomes available, and have the LocationUpdated signal be emitted when location is discovered. And in all of my testing to date, auto-detecting my location as London has always worked, both on VMs (where a "wired" network is available from boot) and on real hardware (where I have to configure WiFi in the FBE). So, what exactly does this PR fix?

gnome-initial-setup/pages/timezone/gis-timezone-page.c
joaquimrocha: Well, nothing really holds a ref to the network monitor here. But I get that it's safer, so if I can do that.

joaquimrocha:
BTW, the map does get UTC by default if no location has been set, but it has, then there'll be a bubble like widget on the map showing the exact timezone.

gnome-initial-setup/pages/timezone/gis-timezone-page.c
```diff
@@ -455,11 +496,18 @@ gis_timezone_page_dispose (GObject *object)
...
 
   g_clear_object (&priv->dtm);
   g_clear_object (&priv->clock);
 
+  if (priv->network_monitor != NULL && priv->network_monitor_handler_id > 0) {
+    g_signal_handler_disconnect (priv->network_monitor,
+                                 priv->network_monitor_handler_id);
+    priv->network_monitor_handler_id = 0;
+  }
+  g_clear_object (&priv->network_monitor);
```
cosimoc: `g_network_monitor_get_default()` does not return a new reference. In fact there should be no need for you to even store it in the private pointer here, since it's already a singleton.

gnome-initial-setup/pages/timezone/gis-timezone-page.c
```diff
@@ -445,6 +478,14 @@ gis_timezone_page_constructed (GObject *object)
   priv->map_location_changed_id =
     g_signal_connect (priv->map, "location-changed",
                       G_CALLBACK (map_location_changed), page);
+  priv->network_monitor_handler_id =
+    g_signal_connect (priv->network_monitor,"network-changed",
```
cosimoc: Missing space after comma

gnome-initial-setup/pages/timezone/gis-timezone-page.c
```diff
@@ -403,6 +410,33 @@ stop_geolocation (GisTimezonePage *page)
...
+      g_signal_handler_disconnect (priv->network_monitor,
+                                   priv->network_monitor_handler_id);
+      priv->network_monitor_handler_id = 0;
+      return;
+    }
+
+  /* if we don't have a geoclue client yet, try to get it and the location,
+   * if we already do, then let's just get the location */
+  if (priv->geoclue_client == NULL)
+    get_location_from_geoclue_async (page);
```
cosimoc: It strikes me that you could theoretically end up in the following situation:
- network is available; from `gis_timezone_page_constructed()` we call `get_location_from_geoclue_async()`
- network changes not to be available. This callback is called and immediately returns
- network changes to available again. This callback calls again `get_location_from_geoclue_async()` since the first geoclue async call hasn't returned yet and we haven't finished constructing the geoclue client yet

This will lead to the creation of two geoclue client objects, which is not what we want. AFAICS you can avoid that by calling `stop_geolocatiion_if_needed()` in this callback, when `network_available` is not `TRUE`.

joaquimrocha: I think it may actually be safer to stop any ongoing geoclue tasks in the `get_location_from_geoclue_async`.

gnome-initial-setup/pages/timezone/gis-timezone-page.c
```diff
@@ -454,12 +494,18 @@ gis_timezone_page_dispose (GObject *object)
...
   GisTimezonePagePrivate *priv = gis_timezone_page_get_instance_private (page);
+  GNetworkMonitor *monitor = g_network_monitor_get_default ();
 
-  stop_geolocation (page);
+  stop_geolocation_if_needed (page);
 
   g_clear_object (&priv->dtm);
   g_clear_object (&priv->clock);
 
+  if (monitor != NULL && priv->network_monitor_handler_id > 0) {
```
cosimoc: Is `monitor` ever going to be `NULL`? The other hunk of this patch assumes it won't.

joaquimrocha: You're right. This was left over after I stopped using the monitor as a private member. I will change it.

joaquimrocha:
Updated the PR with a fixup commit.

joaquimrocha:
Closing as invalid. See discussion in the task.

PR #248, opened 2018-08-08
T23524

joaquimrocha:
https://phabricator.endlessm.com/T23524


joaquimrocha:
Note it's only the last commit (`Show a placeholder if the page needs time...`) that should be reviewed here.

wjt:
We took a different approach which discourages the user from continuing to the timezone page until Geoclue is likely to have had time to find the location.

PR #254, opened 2018-10-18
summary: Allow to set "hack" related contents for the summary page

joaquimrocha:
https://phabricator.endlessm.com/T24135


gnome-initial-setup/pages/summary/gis-summary-page.c
```diff
@@ -39,9 +39,15 @@
 
 #define SERVICE_NAME "gdm-password"
 
+#define CONFIG_SUMMARY_GROUP "page.summary"
+#define CONFIG_SUMMARY_CONTENT_MODE_KEY "content-mode"
+#define HACK_CONTENT_MODE "hack"
```
wjt: Maybe this ship sailed a long time ago, but this information (is it a hack image?) is already available without having to go via the config file. We check the eos-image-version xattr for several other product-specific tweaks here: https://github.com/endlessm/gnome-initial-setup/blob/a3bdc8506e9f80eb7a6bb5f61edb353185caa695/gnome-initial-setup/gis-driver.c#L190-L203

gnome-initial-setup/pages/summary/gis-summary-page.c
```diff
@@ -339,6 +346,24 @@ gis_summary_page_constructed (GObject *object)
                             "lost forever when you restart or shutdown the computer.</b>"));
   }
 
+  /* Override the contents if needed */
+  summary_mode = gis_driver_conf_get_string (GIS_PAGE (page)->driver,
```
wjt: Have you tested whether the text gets updated when you change translation on the first page? (I accept that the live session stuff also makes the same assumption that the page will be torn down and rebuilt if the locale changes.) Otherwise at least some of this should live in locale_changed below.

gnome-initial-setup/pages/summary/gis-summary-page.ui
```diff
@@ -19,12 +19,81 @@
...
+                            <property name="visible">True</property>
+                            <property name="can_focus">False</property>
+                            <property name="icon-name">dialog-warning-symbolic</property>
+                            <property name="pixel_size">24</property>
+                            <property name="valign">center</property>
+                            <property name="hexpand">True</property>
+                          </object>
+                        </child>
+                        <child>
+                          <object class="GtkLabel" id="tagline">
```
wjt: Have you tested how the "all data will be lost" warning looks in live + hack and live + not-hack modes? It looks like you're moving it above the "Start using Endless" button. Why?

gnome-initial-setup/pages/summary/gis-summary-page.c
```diff
@@ -339,6 +346,24 @@ gis_summary_page_constructed (GObject *object)
...
+                                             CONFIG_SUMMARY_GROUP,
+                                             CONFIG_SUMMARY_CONTENT_MODE_KEY);
+  if (g_strcmp0 (summary_mode, HACK_CONTENT_MODE) == 0) {
+    gtk_orientable_set_orientation (GTK_ORIENTABLE (priv->summary_contents_box),
+                                    GTK_ORIENTATION_HORIZONTAL);
+
+    gtk_image_set_from_resource (priv->image, "/org/gnome/initial-setup/ada.png");
```
wjt: Personally I'd prefer to have this layout specified in the `.ui` file in a hidden-by-default widget and have the C code just toggle visibility depending on whether we're in hack mode or not. This is what I did for switching USB and DVD assets and I think that is neater. In theory it would mean you could preview/tweak the layout in Glade without having to rebuild and step through the FBE to test every change.

https://github.com/endlessm/gnome-initial-setup/blob/a3bdc8506e9f80eb7a6bb5f61edb353185caa695/gnome-initial-setup/pages/live-chooser/gis-live-chooser-page.ui#L76-L99

https://github.com/endlessm/gnome-initial-setup/blob/a3bdc8506e9f80eb7a6bb5f61edb353185caa695/gnome-initial-setup/pages/live-chooser/gis-live-chooser-page.c#L164-L167

gnome-initial-setup/pages/summary/gis-summary-page.ui
```diff
@@ -19,12 +19,81 @@
...
-                <property name="height_request">96</property>
+                <property name="halign">center</property>
+                <property name="valign">center</property>
+                <property name="orientation">vertical</property>
+                <property name="spacing">12</property>
+                <child type="center">
+                  <object class="GtkImage" id="image">
+                    <property name="visible">True</property>
+                    <property name="resource">/org/gnome/initial-setup/ready-to-go.svg</property>
+                    <property name="width_request">96</property>
```
wjt: The ada.png asset is 350x350 – is the size it comes out as correct?

gnome-initial-setup/pages/summary/gis-summary-page.c
joaquimrocha: I have used the image attribute before, but since we have been using the conf file, I thought it'd be more flexible to do so. In any case, I didn't notice you had an image check already in g-i-s, so I can change this. Thanks.

gnome-initial-setup/pages/summary/gis-summary-page.c
joaquimrocha: I haven't. It's important that translations work, that's why I didn't want the text added to the conf file, but it's not super important if in Hack images that bit is not translated (as we're supporting only English for now).

gnome-initial-setup/pages/summary/gis-summary-page.ui
joaquimrocha: I did! I think I had actually seen it above the button, but maybe I am wrong. I will double check.

gnome-initial-setup/pages/summary/gis-summary-page.c
joaquimrocha: It's seems a bit overkill to me to do that when we have already an image + text showing. But I can still do it..

gnome-initial-setup/pages/summary/gis-summary-page.ui
joaquimrocha: Yeah. It takes the images' size, so it looks fine.

gnome-initial-setup/pages/summary/gis-summary-page.c
wjt: I think this approach of changing the orientation and messing with the size requests is more error-prone than defining two sets of widgets. I could be wrong…

gnome-initial-setup/pages/summary/gis-summary-page.ui
```diff
@@ -27,6 +27,42 @@
...
+                <property name="visible">False</property>
+                <property name="halign">center</property>
+                <property name="valign">center</property>
+                <property name="margin_left">20</property>
+                <property name="margin_right">20</property>
+                <property name="orientation">horizontal</property>
+                <property name="spacing">12</property>
+                <child>
+                  <object class="GtkImage">
+                    <property name="visible" bind-source="image" bind-property="visible" bind-flags="invert-boolean|sync-create|bidirectional">False</property>
```
wjt: Interesting approach – does this work as expected given it's the visibility of `hack_box` that you're adjusting in the code, not its children?

gnome-initial-setup/gis-driver.c
```diff
@@ -187,6 +189,13 @@ get_image_version (void)
   return image_version;
 }
 
+const gchar *
+gis_driver_get_image_version (GisDriver *driver)
```
wjt: I think this would be better if the driver split the image ID at the first `-`, stored that as `priv->product`, and then had:

`gboolean gis_driver_is_product (GisDriver *, const char *)` for a reason that will become clear…

gnome-initial-setup/pages/summary/gis-summary-page.c
```diff
@@ -339,6 +342,11 @@ gis_summary_page_constructed (GObject *object)
                             "lost forever when you restart or shutdown the computer.</b>"));
   }
 
+  /* Show the right contents for Hack images */
+  if (g_str_has_prefix (gis_driver_get_image_version (GIS_PAGE (page)->driver),
```
wjt: This will CRITICAL if `gis_driver_get_image_version` returns NULL. In practice it will not on any Endless system that originated as an OS image, but if you (eg) debootstrap yourself a new dev environment it will be. All other places that check the image version include a NULL check so I think this one should too. That's why I think a better approach would be for driver to "parse" eos-image-version (ie split at `-`) then use `g_strcmp0()`. But if you're in a rush, just add the NULL check here I guess.

gnome-initial-setup/pages/summary/gis-summary-page.ui
wjt: Interesting approach – does this work as expected given it's the visibility of `hack_box` that you're adjusting in the code, not its children?

gnome-initial-setup/pages/summary/gis-summary-page.c
joaquimrocha: Yeah. I guess it's better to account for `NULL` even though it's unlikely.

joaquimrocha: Yeah. I guess it's better to account for `NULL` even though it's unlikely.

gnome-initial-setup/pages/summary/gis-summary-page.ui
joaquimrocha: I use `show_all`, that's why it works.

joaquimrocha:
Updated the commits. Thanks for another review @wjt .

PR #255, opened 2018-10-25
summary: add support to launch initial contact app on Hack images

cosimoc:
On hack images, upon the completion of the FBE, we want to launch our
"HackUnlock" puzzle application. We use Clippy to connect to the state
notification for its unlocked state; when that is received we can
proceed with the login as usual.

Note that due to a race condition in when Clippy exports its
interface, we need to add a timeout before calling the Connect()
method. This is possible to fix with a different Clippy/GTK
integration, or a different Clippy API, but for now the timeout at
least allows us to move forward without blocking on that.

https://phabricator.endlessm.com/T23942

### DIFF
--- a/gnome-initial-setup/pages/summary/gis-summary-page.c
+++ b/gnome-initial-setup/pages/summary/gis-summary-page.c
@@ -51,6 +51,10 @@ struct _GisSummaryPagePrivate {
 
   ActUser *user_account;
   const gchar *user_password;
+
+  GDBusProxy *clippy_proxy;
+  gulong clippy_proxy_signal_id;
+  guint clippy_timeout_id;
 };
 typedef struct _GisSummaryPagePrivate GisSummaryPagePrivate;
 
@@ -219,7 +223,7 @@ log_user_in (GisSummaryPage *page)
 }
 
 static void
-done_cb (GtkButton *button, GisSummaryPage *page)
+finish_fbe (GisSummaryPage *page)
 {
   gis_ensure_stamp_files ();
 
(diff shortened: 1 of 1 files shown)


gnome-initial-setup/pages/summary/gis-summary-page.c
```diff
@@ -236,6 +239,117 @@ done_cb (GtkButton *button, GisSummaryPage *page)
...
+    g_signal_connect (priv->clippy_proxy, "g-signal",
+                      G_CALLBACK (initial_contact_app_signal), page);
+}
+
+static gboolean
+initial_contact_connect_to_clippy_timeout (gpointer data)
+{
+  GisSummaryPage *page = data;
+  initial_contact_connect_to_clippy (page);
+  return G_SOURCE_REMOVE;
```
joaquimrocha: I wonder if we should just rely on the 1 second timeout for this. Maybe a more robust strategy would be to keep trying for at least 3 times (3 secs). But eventually we'll fix the delay issue, so it's fine either way.

gnome-initial-setup/pages/summary/gis-summary-page.c
```diff
@@ -349,6 +463,22 @@ gis_summary_page_constructed (GObject *object)
...
+static void
+gis_summary_page_finalize (GObject *object)
+{
+  GisSummaryPage *page = GIS_SUMMARY_PAGE (object);
+  GisSummaryPagePrivate *priv = gis_summary_page_get_instance_private (page);
+
+  if (priv->clippy_proxy)
+    {
+      g_signal_handler_disconnect (priv->clippy_proxy,
+                                   priv->clippy_proxy_signal_id);
```
joaquimrocha: Since we only have one reference on the object, do you need to disconnect directly? I guess it's okay to be defensive in any case...

gnome-initial-setup/pages/summary/gis-summary-page.c
```diff
@@ -236,6 +239,117 @@ done_cb (GtkButton *button, GisSummaryPage *page)
...
+}
+
+static void
+initial_contact_connect_to_clippy (GisSummaryPage *page)
+{
+  GisSummaryPagePrivate *priv = gis_summary_page_get_instance_private (page);
+  g_autoptr(GError) error = NULL;
+
+  g_dbus_proxy_call_sync (priv->clippy_proxy,
+                          "Connect",
```
joaquimrocha: Do we need to call this? I thought we'd get the `ObjectNotify` signal in any case when it's changed from the app.

gnome-initial-setup/pages/summary/gis-summary-page.c
wjt: What if the proxy has internal references due to calls which are in flight? I think it's right to do this.

gnome-initial-setup/pages/summary/gis-summary-page.c
```diff
@@ -236,6 +239,117 @@ done_cb (GtkButton *button, GisSummaryPage *page)
...
+  if (error != NULL)
+    {
+      g_critical ("Could not call Connect on DBus proxy for HackUnlock App: %s", error->message);
+
+      finish_fbe (page);
+      return;
+    }
+
+  priv->clippy_proxy_signal_id =
+    g_signal_connect (priv->clippy_proxy, "g-signal",
```
wjt: I was surprised to learn that the D-Bus signal name is not available as the detail, but I guess that's because details get interned with GQuark so this would cause unbounded memory use.

gnome-initial-setup/pages/summary/gis-summary-page.c
```diff
@@ -236,6 +239,117 @@ done_cb (GtkButton *button, GisSummaryPage *page)
...
+{
+  GisSummaryPage *page = user_data;
+  GisSummaryPagePrivate *priv = gis_summary_page_get_instance_private (page);
+  const gchar *object, *property;
+  g_autoptr(GVariant) value;
+
+  if (g_strcmp0 (signal_name, "ObjectNotify") != 0)
+    return;
+
+  g_variant_get (parameters, "(&s&sv)", &object, &property, &value);
```
wjt: Since you go to the trouble below of type-checking `value`, I think it's worth type-checking `parameters` here. (If it has the wrong type you'll get criticals from deep inside GVariant, and then `object`, `property` (and `value` as pointed out above) will have undefined contents and you'll be sad.

gnome-initial-setup/pages/summary/gis-summary-page.c
```diff
@@ -236,6 +239,117 @@ done_cb (GtkButton *button, GisSummaryPage *page)
...
+    {
+      g_critical ("Could not get DBus proxy for HackUnlock App: %s", error->message);
+
+      finish_fbe (page);
+      return;
+    }
+
+  priv->clippy_proxy = g_steal_pointer (&proxy);
+
+  /* FIXME: we need to add a timeout because the Clippy interface is not available yet... */
```
wjt: This is nasty! Could we at least link to the Phabricator ticket tracking this problem?

gnome-initial-setup/pages/summary/gis-summary-page.c
wjt: However, it's possible for `clippy_proxy` to be non-`NULL` but `priv->clippy_proxy_signal_id` to be `0` in which case this call will critical.

gnome-initial-setup/pages/summary/gis-summary-page.c
```diff
@@ -236,6 +239,117 @@ done_cb (GtkButton *button, GisSummaryPage *page)
...
+  g_timeout_add_seconds (1, initial_contact_connect_to_clippy_timeout, page);
```
wjt: Take a ref to `page` & release it (either with _full or in the callback)?

gnome-initial-setup/pages/summary/gis-summary-page.c
joaquimrocha: @wjt There's [this](https://phabricator.endlessm.com/T24047). It can be tracked in it or edited to be more about Clippy and less about the quests stuff.

cosimoc:
Thanks @wjt and @joaquimrocha. I addressed the comments.

gnome-initial-setup/pages/summary/gis-summary-page.c
wjt: Stupid question that I can't believe I don't know the answer to: if the callback returns `G_SOURCE_REMOVE`, is the ID returned by `g_timeout_add` still valid?

cosimoc: It isn't! I need to reset it here. Good catch!

PR #257, opened 2018-11-01
account: add build flag and env variable to override faces dir

dnohales:
https://phabricator.endlessm.com/T23766

### DIFF
--- a/gnome-initial-setup/pages/account/um-photo-dialog.c
+++ b/gnome-initial-setup/pages/account/um-photo-dialog.c
@@ -169,13 +169,40 @@ menu_item_for_filename (UmPhotoDialog *um,
         return menuitem;
 }
 
+static GSList *
+get_facesdirs (void)
+{
+        GSList *facesdirs = NULL;
+        const gchar * const * data_dirs;
+        int i;
+        const char *facesdir_env;
+
+        facesdir_env = g_getenv ("GIS_ACCOUNT_PAGE_FACESDIR");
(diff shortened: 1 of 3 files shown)


gnome-initial-setup/pages/account/um-photo-dialog.c
```diff
@@ -169,13 +169,40 @@ menu_item_for_filename (UmPhotoDialog *um,
...
+                                             g_strdup (ACCOUNT_PAGE_FACESDIR));
+        }
+
+        data_dirs = g_get_system_data_dirs ();
+        for (i = 0; data_dirs[i] != NULL; i++) {
+                char *path = g_build_filename (data_dirs[i], "pixmaps", "faces", NULL);
+                facesdirs = g_slist_prepend (facesdirs, path);
+        }
+
+        return g_slist_reverse (facesdirs);
```
wjt: If you build the list with `GQueue` (which maintains a tail pointer) you don't need to do the "prepend then reverse" dance.

cosimoc: Unless you feel strongly about this, I will leave it with a list for now since it's just a few items.

wjt: No, I don't feel strongly about it. (I have felt weakly but persistently about it since 2011, when I wrote a blog post about it that was lost in a database crash, but that didn't stop “Bob Bobson” [critiquing it](https://blogs.gnome.org/wjjt/2011/03/13/emusic-in-banshee/#comment-14).)

PR #264, opened 2018-11-08
account: allow to specify a different avatar from vendor conf

cosimoc:
This will allow images to customize the default avatar.

https://phabricator.endlessm.com/T24421

### DIFF
--- a/gnome-initial-setup/pages/account/gis-account-page-local.c
+++ b/gnome-initial-setup/pages/account/gis-account-page-local.c
@@ -44,6 +44,9 @@
 #define SHARED_ACCOUNT_USERNAME "shared"
 #define SHARED_ACCOUNT_FULLNAME "Shared Account"
 
+#define CONFIG_ACCOUNT_GROUP "page.account"
+#define CONFIG_ACCOUNT_DEFAULT_AVATAR_KEY "default-avatar"
+
 struct _GisAccountPageLocalPrivate
 {
   GtkWidget *avatar_button;
@@ -388,11 +391,32 @@ confirm (GisAccountPageLocal *page)
     g_signal_emit (page, signals[CONFIRM], 0);
 }
 
+static gchar *
+get_default_avatar_image (void)
+{
+  GisDriver *driver;
+  gchar *default_avatar;
+
+  driver = GIS_DRIVER (g_application_get_default ());
+  default_avatar = gis_driver_conf_get_string (driver,
+                                               CONFIG_ACCOUNT_GROUP,
+                                               CONFIG_ACCOUNT_DEFAULT_AVATAR_KEY);
+
+  if (default_avatar && !g_file_test (default_avatar, G_FILE_TEST_EXISTS))
+    g_clear_pointer (&default_avatar, g_free);
+
+  if (!default_avatar)
+    default_avatar = g_strdup (AVATAR_IMAGE_DEFAULT);
+
+  return default_avatar;
+}
+
 static void
 gis_account_page_local_constructed (GObject *object)
 {
   GisAccountPageLocal *page = GIS_ACCOUNT_PAGE_LOCAL (object);
   GisAccountPageLocalPrivate *priv = gis_account_page_local_get_instance_private (page);
+  g_autofree gchar *default_avatar = NULL;
 
   G_OBJECT_CLASS (gis_account_page_local_parent_class)->constructed (object);
 
@@ -442,7 +466,8 @@ gis_account_page_local_constructed (GObject *object)
                                             avatar_callback,
                                             page);
 
-  avatar_callback (NULL, AVATAR_IMAGE_DEFAULT, page);
+  default_avatar = get_default_avatar_image ();
+  avatar_callback (NULL, default_avatar, page);
 
   validate (page);
 }
--- a/gnome-initial-setup/pages/account/um-photo-dialog.c
+++ b/gnome-initial-setup/pages/account/um-photo-dialog.c
@@ -41,8 +41,8 @@
 
 #define ROW_SPAN 6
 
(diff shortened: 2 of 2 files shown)


erikos:
Overall, looks good to me, maybe Joaquim can take a look as well.

gnome-initial-setup/pages/account/gis-account-page-local.c
```diff
@@ -44,6 +44,9 @@
 #define SHARED_ACCOUNT_USERNAME "shared"
 #define SHARED_ACCOUNT_FULLNAME "Shared Account"
 
+#define VENDOR_ACCOUNT_GROUP "Account"
```
joaquimrocha: I think this should be `page.account` as is the convention we've been using since our previous changes in g-i-s this year.

gnome-initial-setup/pages/account/gis-account-page-local.c
```diff
@@ -44,6 +44,9 @@
...
+#define VENDOR_ACCOUNT_DEFAULT_AVATAR_KEY "Default Avatar"
```
joaquimrocha: And this should be `default-avatar`.

gnome-initial-setup/pages/account/gis-account-page-local.c
cosimoc: That doesn't seem right; in `cc-common-language.c` we use `Language` as the group name [1], and in `gis-branding-welcome-page.c` we use `Welcome` [2].
We also already committed code that uses this group name for the account page: see [3].

[1] https://github.com/endlessm/gnome-initial-setup/blob/master/gnome-initial-setup/pages/language/cc-common-language.c#L290
[2] https://github.com/endlessm/gnome-initial-setup/blob/master/gnome-initial-setup/pages/branding-welcome/gis-branding-welcome-page.c#L49
[3] https://github.com/endlessm/gnome-initial-setup/commit/f13afa435cd70e3e7e95a4601109a721b049c10a

gnome-initial-setup/pages/account/gis-account-page-local.c
cosimoc: In the code we committed to add the extra "faces" directory, we called the key `Faces` (still part of this commit: https://github.com/endlessm/gnome-initial-setup/commit/f13afa435cd70e3e7e95a4601109a721b049c10a).
Does that mean that I should also rename that?

gnome-initial-setup/pages/account/gis-account-page-local.c
joaquimrocha: I think we had this discussion before (I cannot remember if it was with you), but we had settled that using `page.id-of-page` is better than having just a capitalized name.
We have this code in place already too:
https://github.com/endlessm/gnome-initial-setup/blob/master/gnome-initial-setup/pages/account/gis-account-page.c#L35
https://github.com/endlessm/gnome-initial-setup/blob/master/gnome-initial-setup/pages/timezone/gis-timezone-page.c#L60
(the latter is also upstream)

So I guess it's update to you but it'd be nice if they all converge. Maybe @wjt can untie this.

gnome-initial-setup/pages/account/gis-account-page-local.c
joaquimrocha: I think we should. But like in my previous comment, maybe Will can untie this.

cosimoc:
There is really nothing for Will to “untie”... I want to do things in the expected way and don’t feel strongly about which convention we use myself. I was only pointing out things aren’t consistent at the moment.
I’ll rework them to be consistent as part of this PR.

gnome-initial-setup/pages/account/gis-account-page-local.c
```diff
@@ -388,11 +391,32 @@ confirm (GisAccountPageLocal *page)
     g_signal_emit (page, signals[CONFIRM], 0);
 }
 
+static gchar *
+get_default_avatar_image (void)
+{
+  GisDriver *driver;
+  gchar *default_avatar;
+
+  driver = GIS_DRIVER (g_application_get_default ());
```
wjt:
```suggestion
  driver = GIS_PAGE (self)->driver;
```

is a bit less obscure.

gnome-initial-setup/pages/account/gis-account-page-local.c
wjt: Here is the previous discussion. https://github.com/endlessm/gnome-initial-setup/pull/239#discussion_r204589418

wjt: (As explained there, the non-namespaced group names were already present when the `page.` convention was introduced, and we must support them forever because we can't update the vendor conf file.)

cosimoc: Sounds good. As part of this PR, I will make sure the new options we added for the account page (the faces dir and the default avatar) use the `page.account` convention. I won't touch the other places in g-i-s where the older convention is used.

gnome-initial-setup/pages/account/gis-account-page-local.c
cosimoc: Unfortunately I can't do that. This is `GisAccountPageLocal` which is not a `GisPage`.

wjt: Good point, well made.

joaquimrocha:
> There is really nothing for Will to “untie”... I want to do things in the expected way and don’t feel strongly about which convention we use myself. I was only pointing out things aren’t consistent at the moment.
> I’ll rework them to be consistent as part of this PR.

Thanks Cosimo. Maybe "untie" was the wrong verb, I just wanted to make sure Will validated the right direction here.

PR #269, opened 2018-11-27
Connect to GDM as soon as gnome-initial-setup starts

jprvita:
GDM can't tell Plymouth to quit until gnome-initial-setup connects to
it, so GDM knows gnome-initial-setup has started successfully. Currently
gnome-initial-setup only connects to GDM once the user clicks the "Start
Using $DISTRO" button on the last page, to switch over to the
newly-created user. This prevents accessing the TTY for the whole time
gnome-initial-setup is running and makes systemd account the time spent
in gnome-initial-setup as part of the boot process.

Connecting to GDM as soon as gnome-initial-setup starts allows GDM to
tell Plymouth to quit at the right time, avoiding the problems described
above.

https://phabricator.endlessm.com/T21181

### DIFF
--- a/gnome-initial-setup/pages/summary/gis-summary-page.c
+++ b/gnome-initial-setup/pages/summary/gis-summary-page.c
@@ -78,6 +78,7 @@ connect_to_gdm (GdmGreeter      **greeter,
     g_error_free (error);
   }
 
+  g_clear_object (&client);
   return;
 }
 
@@ -176,7 +177,7 @@ static void
 log_user_in (GisSummaryPage *page)
 {
   GisSummaryPagePrivate *priv = gis_summary_page_get_instance_private (page);
-  GError *error = NULL;
+  g_autoptr(GError) error = NULL;
 
   if (!priv->greeter || !priv->user_verifier) {
     g_warning ("No GDM connection; not initiating login");
--- a/gnome-initial-setup/pages/summary/gis-summary-page.c
+++ b/gnome-initial-setup/pages/summary/gis-summary-page.c
@@ -50,6 +50,9 @@ struct _GisSummaryPagePrivate {
   ActUser *user_account;
   const gchar *user_password;
 
+  GdmGreeter *greeter;
+  GdmUserVerifier *user_verifier;
+
   GDBusProxy *clippy_proxy;
   gulong clippy_proxy_signal_id;
 };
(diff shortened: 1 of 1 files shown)


gnome-initial-setup/gnome-initial-setup.c
```diff
@@ -286,6 +287,25 @@ get_mode (void)
     return GIS_DRIVER_MODE_NEW_USER;
 }
 
+static gboolean
+connect_to_gdm (void)
+{
+  GdmClient *client;
```
wjt:
```suggestion
  g_autoptr(GdmClient) client = NULL;
```

(if that compiles, else unref on the way out)

gnome-initial-setup/gnome-initial-setup.c
```diff
@@ -286,6 +287,25 @@ get_mode (void)
...
+  GdmGreeter *greeter;
```
wjt:
```suggestion
  g_autoptr(GdmGreeter) greeter = NULL;
```

ditto

gnome-initial-setup/gnome-initial-setup.c
```diff
@@ -286,6 +287,25 @@ get_mode (void)
...
+  GError *error = NULL;
```
wjt:
```suggestion
  g_autoptr(GError) error = NULL;
```

and remove the `g_error_free` call

gnome-initial-setup/gnome-initial-setup.c
```diff
@@ -339,10 +359,16 @@ main (int argc, char *argv[])
...
    * the keyring manually so that we can pass the credentials
    * along to the new user in the handoff.
+   * We also need to connect to GDM so it instructs plymouth to quit,
+   * otherwise it stays around until we reach the last (summary) page.
    */
-  if (mode == GIS_DRIVER_MODE_NEW_USER)
+  if (mode == GIS_DRIVER_MODE_NEW_USER) {
     gis_ensure_login_keyring ();
 
+    if (!connect_to_gdm ())
```
wjt: I don't think this is the right place to do this. The `summary` page will ultimately connect to GDM, so I think it'd be better to move the call to its `connect_to_gdm` function into `gis_summary_page_constructed`, and stash the `GdmGreeter` and `GdmUserVerifier` in its `Private` structure.

jprvita: That makes much more sense. I'm going with that approach (which invalidates the previous comments).

gnome-initial-setup/pages/summary/gis-summary-page.c
```diff
@@ -50,21 +50,22 @@ struct _GisSummaryPagePrivate {
...
 typedef struct _GisSummaryPagePrivate GisSummaryPagePrivate;
 
 G_DEFINE_TYPE_WITH_PRIVATE (GisSummaryPage, gis_summary_page, GIS_TYPE_PAGE);
 
-static gboolean
+static void
 connect_to_gdm (GdmGreeter      **greeter,
                 GdmUserVerifier **user_verifier)
 {
   GdmClient *client;
```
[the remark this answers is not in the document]
jprvita: Using `g_autoptr` here fails to build with
```
../gnome-initial-setup/pages/summary/gis-summary-page.c:67:13: error: unknown type name ‘GdmClient_autoptr’
   g_autoptr(GdmClient) client = NULL;
             ^
```
even though we have the following in our libgdm.h:
```
#if GLIB_CHECK_VERSION(2, 44, 0)
G_DEFINE_AUTOPTR_CLEANUP_FUNC (GdmClient, g_object_unref)                                                                               
#endif
```
So I'm simply calling `g_clear_object (&client)` before exiting this function).

I'm not chaging `g_error_free` to `g_autoptr` where it is not leaking as I don't see the appeal of carrying this type of change downstream. I can add it once I send all of this upstream though.

PR #270, opened 2018-11-27
account: don't dim avatar image preview

cosimoc:
This page will be redesigned in the next rebase, but this is a
high-visibility fix that is at the same time very simple.

https://phabricator.endlessm.com/T24488


cosimoc:
@wjt Thanks, updated.

PR #271, opened 2018-11-27
T23646 eula

dnohales:
https://phabricator.endlessm.com/T23646

### DIFF
--- a/gnome-initial-setup/pages/account/gis-account-page-eula-dialog.c
+++ b/gnome-initial-setup/pages/account/gis-account-page-eula-dialog.c
@@ -1,0 +1,65 @@
+/* -*- mode: C; c-file-style: "gnu"; indent-tabs-mode: nil; -*- */
+/*
+ * Copyright (C) 2018 Endless Mobile, Inc.
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License as
+ * published by the Free Software Foundation; either version 2 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful, but
+ * WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA
+ * 02111-1307, USA.
+ *
+ * Written by:
+ *     Damián Nohales <damian@endlessm.com>
+ */
+
+#include "config.h"
+#include "gis-account-page-eula-dialog.h"
+#include "../endless-eula/gis-endless-eula-viewer.h"
+
+struct _GisAccountPageEulaDialog {
+  GtkDialog parent;
+
+  GtkWidget *eula_viewer_container;
+};
+
+G_DEFINE_TYPE (GisAccountPageEulaDialog, gis_account_page_eula_dialog, GTK_TYPE_DIALOG)
+
+static void
+gis_account_page_eula_dialog_constructed (GObject *object)
+{
+  GisAccountPageEulaDialog *dialog = GIS_ACCOUNT_PAGE_EULA_DIALOG (object);
+  GtkWidget *eula_viewer;
+
+  G_OBJECT_CLASS (gis_account_page_eula_dialog_parent_class)->constructed (object);
+
+  eula_viewer = g_object_new (GIS_TYPE_ENDLESS_EULA_VIEWER, NULL);
+  gtk_container_add (GTK_CONTAINER (dialog->eula_viewer_container), eula_viewer);
+
+  gtk_widget_show_all (GTK_WIDGET (dialog));
+}
+
+static void
+gis_account_page_eula_dialog_class_init (GisAccountPageEulaDialogClass *klass)
+{
+  GObjectClass *object_class = G_OBJECT_CLASS (klass);
+
+  gtk_widget_class_set_template_from_resource (GTK_WIDGET_CLASS (klass), "/org/gnome/initial-setup/gis-account-page-eula-dialog.ui");
+  gtk_widget_class_bind_template_child (GTK_WIDGET_CLASS (klass), GisAccountPageEulaDialog, eula_viewer_container);
+
+  object_class->constructed = gis_account_page_eula_dialog_constructed;
+}
+
+static void
+gis_account_page_eula_dialog_init (GisAccountPageEulaDialog *dialog)
+{
+  gtk_widget_init_template (GTK_WIDGET (dialog));
+}
--- a/gnome-initial-setup/pages/account/gis-account-page-eula-dialog.h
+++ b/gnome-initial-setup/pages/account/gis-account-page-eula-dialog.h
@@ -1,0 +1,36 @@
+/* -*- mode: C; c-file-style: "gnu"; indent-tabs-mode: nil; -*- */
+/*
+ * Copyright (C) 2018 Endless Mobile, Inc
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License as
+ * published by the Free Software Foundation; either version 2 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful, but
+ * WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA
+ * 02111-1307, USA.
+ *
+ * Written by:
+ *     Damián Nohales <damian@endlessm.com>
+ */
+
+#ifndef __GIS_ACCOUNT_PAGE_EULA_DIALOG_H__
+#define __GIS_ACCOUNT_PAGE_EULA_DIALOG_H__
+
+#include "gnome-initial-setup.h"
+
+G_BEGIN_DECLS
+
+#define GIS_TYPE_ACCOUNT_PAGE_EULA_DIALOG (gis_account_page_eula_dialog_get_type ())
+G_DECLARE_FINAL_TYPE (GisAccountPageEulaDialog, gis_account_page_eula_dialog, GIS, ACCOUNT_PAGE_EULA_DIALOG, GtkDialog)
+
+G_END_DECLS
+
+#endif /* __GIS_ACCOUNT_PAGE_EULA_DIALOG_H__ */
--- a/gnome-initial-setup/pages/account/gis-account-page-local.c
+++ b/gnome-initial-setup/pages/account/gis-account-page-local.c
@@ -22,7 +22,9 @@
 #include "config.h"
 
 #include "gis-page.h"
+#include "gis-account-page.h"
 #include "gis-account-page-local.h"
+#include "gis-account-page-eula-dialog.h"
 #include "gnome-initial-setup.h"
 
 #include <glib/gi18n.h>
@@ -57,6 +59,7 @@ struct _GisAccountPageLocalPrivate
   gboolean   has_custom_username;
   GtkWidget *username_explanation;
   GtkWidget *password_switch;
+  GtkWidget *eulanote_label;
   gboolean passwordless;
   UmPhotoDialog *photo_dialog;
 
@@ -417,11 +420,27 @@ get_default_avatar_image (void)
   return default_avatar;
 }
 
+static gboolean
+show_eula_dialog (GtkLabel    *label,
+                  const gchar *uri,
+                  gpointer     user_data)
+{
+  GtkWindow *toplevel = GTK_WINDOW (gtk_widget_get_toplevel (GTK_WIDGET (label)));
+  GisAccountPageEulaDialog *dialog = g_object_new (GIS_TYPE_ACCOUNT_PAGE_EULA_DIALOG,
+                                                   "use-header-bar", TRUE,
+                                                   "transient-for", toplevel,
+                                                   NULL);
+  g_signal_connect (dialog, "response",
+                    G_CALLBACK (gtk_widget_destroy), NULL);
+  return TRUE;
+}
+
 static void
 gis_account_page_local_constructed (GObject *object)
 {
   GisAccountPageLocal *page = GIS_ACCOUNT_PAGE_LOCAL (object);
   GisAccountPageLocalPrivate *priv = gis_account_page_local_get_instance_private (page);
+  GisDriver *driver;
   g_autofree gchar *default_avatar = NULL;
 
   G_OBJECT_CLASS (gis_account_page_local_parent_class)->constructed (object);
@@ -444,6 +463,8 @@ gis_account_page_local_constructed (GObject *object)
                             G_CALLBACK (confirm), page);
   g_signal_connect_swapped (priv->password_switch, "notify::active",
                             G_CALLBACK (validate), page);
+  g_signal_connect (priv->eulanote_label, "activate-link",
+                    G_CALLBACK (show_eula_dialog), page);
 
   priv->valid_name = FALSE;
   priv->valid_username = FALSE;
@@ -476,6 +497,10 @@ gis_account_page_local_constructed (GObject *object)
   avatar_callback (NULL, default_avatar, page);
 
   validate (page);
+
+  driver = GIS_DRIVER (g_application_get_default ());
+  if (gis_driver_is_hack (driver))
+    gtk_widget_show (priv->eulanote_label);
 }
 
 static void
@@ -604,6 +629,7 @@ gis_account_page_local_class_init (GisAccountPageLocalClass *klass)
   gtk_widget_class_bind_template_child_private (GTK_WIDGET_CLASS (klass), GisAccountPageLocal, username_combo);
   gtk_widget_class_bind_template_child_private (GTK_WIDGET_CLASS (klass), GisAccountPageLocal, username_explanation);
   gtk_widget_class_bind_template_child_private (GTK_WIDGET_CLASS (klass), GisAccountPageLocal, password_switch);
+  gtk_widget_class_bind_template_child_private (GTK_WIDGET_CLASS (klass), GisAccountPageLocal, eulanote_label);
 
   object_class->constructed = gis_account_page_local_constructed;
   object_class->dispose = gis_account_page_local_dispose;
--- a/gnome-initial-setup/pages/account/gis-account-page-local.c
+++ b/gnome-initial-setup/pages/account/gis-account-page-local.c
@@ -661,7 +661,7 @@ void
 gis_account_page_local_shown (GisAccountPageLocal *local)
 {
   GisAccountPageLocalPrivate *priv = gis_account_page_local_get_instance_private (local);
-  gtk_widget_grab_focus (priv->fullname_entry); 
+  gtk_widget_grab_focus (priv->fullname_entry);
 }
 
 gboolean
--- a/gnome-initial-setup/pages/account/gis-account-page-local.ui
+++ b/gnome-initial-setup/pages/account/gis-account-page-local.ui
@@ -62,6 +62,7 @@
             <property name="column_spacing">12</property>
             <property name="row_spacing">6</property>
             <property name="margin_top">42</property>
+            <property name="halign">center</property>
             <child>
               <object class="GtkLabel" id="fullname_label">
                 <property name="visible">True</property>
@@ -181,6 +182,20 @@
                 <property name="height">1</property>
               </packing>
             </child>
+          </object>
+        </child>
+        <child>
+          <object class="GtkLabel" id="eulanote_label">
+            <property name="visible">False</property>
+            <property name="no_show_all">True</property>
+            <property name="halign">center</property>
+            <property name="valign">end</property>
+            <property name="margin_bottom">16</property>
+            <property name="hexpand">True</property>
+            <property name="vexpand">True</property>
+            <property name="use_markup">True</property>
+            <property name="track_visited_links">False</property>
+            <property name="label" translatable="yes">By creating this account you agree to accept the &lt;a href="#"&gt;Endless Terms of Use&lt;/a&gt;.</property>
           </object>
         </child>
       </object>
--- a/gnome-initial-setup/pages/account/gis-account-page.c
+++ b/gnome-initial-setup/pages/account/gis-account-page.c
@@ -24,6 +24,7 @@
 #define PAGE_ID "account"
 
 #include "config.h"
+#include "gis-page-util.h"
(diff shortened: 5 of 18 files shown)


gnome-initial-setup/pages/account/gis-account-page-eula-dialog.c
```diff
@@ -0,0 +1,319 @@
...
+#include <glib/gi18n.h>
+#include <locale.h>
+#include <webkit2/webkit2.h>
+
+typedef struct {
+  GtkWidget *scrolled_window;
+  GtkWidget *evince_view;
+} GisAccountPageEulaDialogPrivate;
+
+G_DEFINE_TYPE_WITH_PRIVATE (GisAccountPageEulaDialog, gis_account_page_eula_dialog, GTK_TYPE_DIALOG);
```
cosimoc: You can remove the trailing semicolon here

gnome-initial-setup/pages/account/meson.build
```diff
@@ -19,6 +19,8 @@ sources += files(
     'gis-account-page.h',
     'gis-account-pages.c',
     'gis-account-pages.h',
+    'gis-account-page-eula-dialog.h',
+    'gis-account-page-eula-dialog.c',
```
cosimoc: Nitpick: The other files list the .c file before the .h

gnome-initial-setup/pages/account/gis-account-page.c
```diff
@@ -161,12 +162,67 @@ switch_login_mode (GisAccountPage *page)
...
+                                                 NULL,
+                                                 "com.endlessm.Metrics",
+                                                 "/com/endlessm/Metrics",
+                                                 "com.endlessm.Metrics.EventRecorderServer",
+                                                 NULL, &error);
+
+  if (error != NULL)
+    {
+      g_critical ("Unable to create a DBus proxy for the metrics daemon: %s", error->message);
+      g_clear_object (&metrics_proxy);
```
cosimoc: If this is reached, this would always be `NULL`

gnome-initial-setup/pages/account/gis-account-page.c
```diff
@@ -161,12 +162,67 @@ switch_login_mode (GisAccountPage *page)
...
+
+  if (demo_mode)
+    {
+      recorder = emtr_event_recorder_get_default ();
+
+      g_variant_dict_init (&dict, NULL);
+
+      emtr_event_recorder_record_event (recorder,
+                                        DEMO_MODE_ENTERED_METRIC,
+                                        g_variant_dict_end (&dict));
```
cosimoc: Could you not pass `NULL` or just an empty variant instead of creating this empty dict here?

gnome-initial-setup/pages/account/gis-account-page.c
```diff
@@ -161,12 +162,67 @@ switch_login_mode (GisAccountPage *page)
...
+    }
+}
+
 static gboolean
 gis_account_page_apply (GisPage *gis_page,
                         GCancellable *cancellable)
 {
   GisAccountPage *page = GIS_ACCOUNT_PAGE (gis_page);
   GisAccountPagePrivate *priv = gis_account_page_get_instance_private (page);
+  gboolean demo_mode = gis_driver_is_in_demo_mode (gis_page->driver);
```
cosimoc: Demo mode is disabled on Hack, but to the best of my recollection, the account page is skipped entirely on demo mode anyway. So I don't think you need to be concerned of reporting the demo mode metric from here.

gnome-initial-setup/pages/account/gis-account-page-local.ui
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <interface>
-  <!-- interface-requires gtk+ 3.0 -->
+  <!-- interface-requires gtk+ 3.30 -->
```
cosimoc: Did you get this change from Glade or did you make it manually? Generally we avoid touching unrelated areas of the code.

gnome-initial-setup/pages/account/gis-account-page-local.c
```diff
@@ -661,7 +682,7 @@ void
 gis_account_page_local_shown (GisAccountPageLocal *local)
 {
   GisAccountPageLocalPrivate *priv = gis_account_page_local_get_instance_private (local);
-  gtk_widget_grab_focus (priv->fullname_entry); 
+  gtk_widget_grab_focus (priv->fullname_entry);
```
cosimoc: I'd remove this unrelated whitespace fix.

gnome-initial-setup/pages/account/gis-account-page-local.c
```diff
@@ -417,6 +419,22 @@ get_default_avatar_image (void)
...
 }
 
+static gboolean
+show_eula_dialog (GtkLabel    *label,
+                  const gchar *uri,
+                  gpointer     user_data)
+{
+  GtkWindow *toplevel = GTK_WINDOW (gtk_widget_get_toplevel (GTK_WIDGET (label)));
+  GisAccountPageEulaDialog *dialog = g_object_new (GIS_TYPE_ACCOUNT_PAGE_EULA_DIALOG,
+                                                   "use-header-bar", 1,
```
cosimoc: Use `TRUE1 instead of `1`

gnome-initial-setup/pages/account/gis-account-page-eula-dialog.c
```diff
@@ -0,0 +1,319 @@
...
+
+static void
+gis_account_page_eula_dialog_constructed (GObject *object)
+{
+  GisAccountPageEulaDialog *dialog = GIS_ACCOUNT_PAGE_EULA_DIALOG (object);
+
+  G_OBJECT_CLASS (gis_account_page_eula_dialog_parent_class)->constructed (object);
+
+  load_css_overrides (dialog);
+  gtk_widget_show_all (GTK_WIDGET (dialog));
```
cosimoc: The caller also calls `show_all()` on the dialog. Can probably remove one or the other

gnome-initial-setup/pages/account/gis-account-page-eula-dialog.c
```diff
@@ -0,0 +1,319 @@
...
+
+  g_object_unref (provider);
+}
+
+static void
+gis_account_page_eula_dialog_constructed (GObject *object)
+{
+  GisAccountPageEulaDialog *dialog = GIS_ACCOUNT_PAGE_EULA_DIALOG (object);
+
+  G_OBJECT_CLASS (gis_account_page_eula_dialog_parent_class)->constructed (object);
```
cosimoc: Does it still work if you just do all of this from _init instead?

dnohales: No, it doesn't work.

dnohales:
Updated.

gnome-initial-setup/pages/account/gis-account-page.c
```diff
@@ -161,13 +163,52 @@ switch_login_mode (GisAccountPage *page)
     set_mode (page, UM_LOCAL);
 }
 
+static void
+enable_endlessm_metrics (void)
```
ramcq: Can we put this function in a common location so that it can be shared between the accounts page and here, rather than essentially copy-pasted (apart from the GtkToggle binding)?

gnome-initial-setup/pages/endless-eula/gis-endless-eula-page.c
```diff
@@ -419,8 +130,12 @@ gis_endless_eula_page_constructed (GObject *object)
 static void
 gis_endless_eula_page_locale_changed (GisPage *page)
 {
+  GisEndlessEulaPage *gis_ee_page = GIS_ENDLESS_EULA_PAGE (page);
```
wjt: I'd call this `self`.

gnome-initial-setup/pages/endless-eula/gis-endless-eula-page.c
```diff
@@ -419,8 +130,12 @@ gis_endless_eula_page_constructed (GObject *object)
...
+  GisEndlessEulaPagePrivate *priv = gis_endless_eula_page_get_instance_private (gis_ee_page);
+
   gis_page_set_title (page, _("Terms of Use"));
-  reload_terms_view (GIS_ENDLESS_EULA_PAGE (page));
+  if (priv->eula_viewer != NULL)
```
wjt: In what circumstances could this be `NULL`?

gnome-initial-setup/pages/endless-eula/gis-endless-eula-page.c
```diff
@@ -398,14 +109,14 @@ gis_endless_eula_page_constructed (GObject *object)
...
 
-  load_css_overrides (page);
-
   gtk_widget_show (GTK_WIDGET (page));
-  load_terms_view (page);
 
   gis_page_set_forward_text (GIS_PAGE (page), _("_Accept and Continue"));
   gis_page_set_complete (GIS_PAGE (page), TRUE);
 
+  priv->eula_viewer = g_object_new (GIS_TYPE_ENDLESS_EULA_VIEWER, NULL);
```
wjt: This needs to be unref'd in dispose/finalize

gnome-initial-setup/pages/endless-eula/gis-endless-eula-page.ui
```diff
@@ -52,12 +52,11 @@
           </packing>
         </child>
         <child>
-          <object class="GtkScrolledWindow" id="eula_scrolledwin">
+          <object class="GtkAlignment" id="eula_viewer_container">
             <property name="visible">True</property>
             <property name="can_focus">False</property>
             <property name="hexpand">True</property>
             <property name="vexpand">True</property>
-            <property name="shadow_type">in</property>
```
wjt: Do we not like shadows any more?

gnome-initial-setup/pages/endless-eula/gis-endless-eula-viewer.h
```diff
@@ -0,0 +1,58 @@
...
+#ifndef __GIS_ENDLESS_EULA_VIEWER_H__
+#define __GIS_ENDLESS_EULA_VIEWER_H__
+
+#include "gnome-initial-setup.h"
+#include <evince-view.h>
+
+G_BEGIN_DECLS
+
+#define GIS_TYPE_ENDLESS_EULA_VIEWER               (gis_endless_eula_viewer_get_type ())
+#define GIS_ENDLESS_EULA_VIEWER(obj)               (G_TYPE_CHECK_INSTANCE_CAST ((obj), GIS_TYPE_ENDLESS_EULA_VIEWER, GisEndlessEulaViewer))
```
wjt: Please can we use `G_DECLARE_FINAL_TYPE` in new code?

gnome-initial-setup/pages/endless-eula/gis-endless-eula-viewer.c
```diff
@@ -0,0 +1,344 @@
...
+#include <evince-view.h>
+#include <glib/gi18n.h>
+#include <locale.h>
+#include <webkit2/webkit2.h>
+
+typedef struct {
+  GtkWidget *evince_view;
+} GisEndlessEulaViewerPrivate;
+
+G_DEFINE_TYPE_WITH_PRIVATE (GisEndlessEulaViewer, gis_endless_eula_viewer, GTK_TYPE_SCROLLED_WINDOW)
```
wjt: If you use `G_DECLARE_FINAL_TYPE` in the header then you can drop `_WITH_PRIVATE` and change the struct above to:

```c
struct _GisEndlessEulaViewer {
  GtkScrolledWindow parent;

  GtkWidget *evince_view;
};
```

and drop all references to the `Private` structure

gnome-initial-setup/pages/account/gis-account-page-eula-dialog.h
```diff
@@ -0,0 +1,55 @@
...
+
+#ifndef __GIS_ACCOUNT_PAGE_EULA_DIALOG_H__
+#define __GIS_ACCOUNT_PAGE_EULA_DIALOG_H__
+
+#include "gnome-initial-setup.h"
+
+G_BEGIN_DECLS
+
+#define GIS_TYPE_ACCOUNT_PAGE_EULA_DIALOG               (gis_account_page_eula_dialog_get_type ())
+#define GIS_ACCOUNT_PAGE_EULA_DIALOG(obj)               (G_TYPE_CHECK_INSTANCE_CAST ((obj), GIS_TYPE_ACCOUNT_PAGE_EULA_DIALOG, GisAccountPageEulaDialog))
```
wjt: `G_DEFINE_FINAL_TYPE`

gnome-initial-setup/pages/account/gis-account-page-local.c
```diff
@@ -417,6 +420,21 @@ get_default_avatar_image (void)
   return default_avatar;
 }
 
+static gboolean
+show_eula_dialog (GtkLabel    *label,
+                  const gchar *uri,
+                  gpointer     user_data)
+{
+  GtkWindow *toplevel = GTK_WINDOW (gtk_widget_get_toplevel (GTK_WIDGET (label)));
+  GisAccountPageEulaDialog *dialog = g_object_new (GIS_TYPE_ACCOUNT_PAGE_EULA_DIALOG,
```
wjt: Personally I'd prefer if this `g_object_new` lived in a `gis_account_page_eula_dialog_new ()` function defined in that .c/.h but hey ho.

gnome-initial-setup/pages/account/gis-account-page.c
wjt: The place would be `gis-page-util.[ch]` which is full of our downstream functionality that is used in >1 page

gnome-initial-setup/pages/account/gis-account-page.c
```diff
@@ -258,6 +299,18 @@ on_shared_user_created (GtkWidget       *page_local,
     gis_driver_set_user_permissions (GIS_PAGE (page)->driver, user, password);
 }
 
+gboolean
+gis_account_page_is_eula_enabled (void)
```
wjt: `gis_account_page_should_show_eula` seems a better name, since the config key is called `show-eula`.

wjt: This function should take a `GisPage *page` and use `page->driver` to get the driver.

wjt:
I'm not wild about the fact that the settings key controlling whether the account page shows a EULA link is independent to the one which hides the EULA page, but it's not the end of the world.

gnome-initial-setup/pages/endless-eula/gis-endless-eula-page.c
dnohales: I had the case of this being `NULL` when initializing the page. I'll double-check what happened there.

gnome-initial-setup/pages/endless-eula/gis-endless-eula-page.ui
dnohales: This is now a `GtkAlignment`, the shadow is applied in `GisEndlessEulaViewer`.

gnome-initial-setup/pages/endless-eula/gis-endless-eula-viewer.h
dnohales: Absolutely, I was just trying to follow the convention.

dnohales:
Updated.

gnome-initial-setup/pages/endless-eula/gis-endless-eula-page.c
cosimoc: It's going to be destroyed as part of the widget hierarchy as far as I can see, so the code we have here should be enough

gnome-initial-setup/pages/endless-eula/gis-endless-eula-viewer.c
```diff
@@ -142,14 +146,15 @@ get_terms_document (void)
    */
   language = setlocale (LC_MESSAGES, NULL);
   locale_variants = g_get_locale_variants (language);
-  path = find_terms_document_for_languages ((const gchar * const *)locale_variants);
-  g_strfreev (locale_variants);
+  driver = GIS_DRIVER (g_application_get_default ());
 
+  if (gis_driver_is_hack (driver))
+    path = find_terms_document_for_languages ("Hack-Terms-of-Use.pdf",
```
cosimoc: Please call it `Endless-Hack-Terms-of-Use.pdf`. However as things are today, this won't work, since we don't have a separate hack terms of use file. Please also open a PR for eos-license-service to create a symlink at least for now.

gnome-initial-setup/gnome-initial-setup.c
```diff
@@ -98,6 +98,13 @@ should_skip_page (GisDriver    *driver,
 {
   guint i = 0;
 
+  /* check the Hack special case first */
+  if (gis_driver_is_hack (driver)) {
+    if (g_strcmp0 ("eula", page_id) == 0 || g_strcmp0 ("endless_eula", page_id) == 0) {
```
cosimoc: You don't need to worry about the `eula` page, we always disable it.
I'd rather disable the `endless_eula` page in the way we disable the others by doing it directly in `gis_prepare_endless_eula_page()`

gnome-initial-setup/pages/endless-eula/gis-endless-eula-page.c
```diff
@@ -42,45 +43,17 @@ typedef struct {
...
-}
-
 static void
 gis_endless_eula_page_finalize (GObject *object)
 {
   GisEndlessEulaPage *page = GIS_ENDLESS_EULA_PAGE (object);
   GisEndlessEulaPagePrivate *priv = gis_endless_eula_page_get_instance_private (page);
 
-  g_clear_object (&priv->metrics_proxy);
+  g_clear_object (&priv->eula_viewer);
```
cosimoc: This is in the wrong commit, but I don't think it's necessary, as mentioned in the other comment.

gnome-initial-setup/pages/endless-eula/gis-endless-eula-viewer.c
dnohales: It will fallback to the regular `Endless-Terms-of-Use.pdf` file.

dnohales:
Updated.

gnome-initial-setup/pages/endless-eula/gis-endless-eula-viewer.c
dnohales: @ramcq Well, my original approach wasn't hack specific since it played with the `vendor.conf` file parameters instead of deciding stuff based on the value of the product name, `g-i-s` is already too hack specific from the point where we added the `gis_driver_is_hack` function. I don't see how this case is different.

So, I'll change it to be: `$datadir/eos-license-service/terms/$product/Terms-of-Use-$locale.pdf`

dnohales:
@ramcq @cosimoc updated

gnome-initial-setup/pages/endless-eula/gis-endless-eula-viewer.c
```diff
@@ -142,14 +152,14 @@ get_terms_document (void)
...
+  driver = GIS_DRIVER (g_application_get_default ());
+  product_name = gis_driver_get_product_name (driver);
 
-  if (path == NULL)
-    {
-      /* Now use "C" as a fallback. */
-      path = find_terms_document_for_language ("C");
-    }
+  path = find_terms_document_for_languages (product_name, (const gchar * const *)locale_variants);
+  if (path == NULL && g_strcmp0 ("eos", product_name) != 0)
```
cosimoc: You could use `!gis_driver_is_product()` here, but do you really need that extra check? Without it, it just means that if we had an `eos` product but no ToS for `eos`, we'd check twice to make that determination. However that doesn't really happen since there always is an `eos` ToS...

dnohales: Ok, if assuming we always trust that there's going to be an `eos` ToS, I agree.

gnome-initial-setup/pages/endless-eula/gis-endless-eula-viewer.c
```diff
@@ -0,0 +1,356 @@
...
+find_terms_document_for_language (const gchar *product_name,
+                                  const gchar *language)
+{
+  const gchar * const * data_dirs;
+  gchar *path = NULL;
+  g_autofree gchar *filename;
+  gint i;
+
+  data_dirs = g_get_system_data_dirs ();
+  filename = g_strconcat ("Terms-of-Use-", language, ".pdf", NULL);
```
wjt: I know it's a bikeshed, but the nice thing about the current structure is that the variable element (the language) is a single path component, so you don't have to glue together the basename of the file like this. Why not:

```
$data_dir/eos-license-service/terms/$product/$language/Terms-of-Use.pdf
```

but I don't feel overwhelmingly strongly about this.

gnome-initial-setup/pages/endless-eula/gis-endless-eula-viewer.c
```diff
@@ -0,0 +1,356 @@
...
+
+G_DEFINE_TYPE (GisEndlessEulaViewer, gis_endless_eula_viewer, GTK_TYPE_SCROLLED_WINDOW)
+
+#define LICENSE_SERVICE_URI "http://localhost:3010"
+
+static void
+present_view_in_modal (GisEndlessEulaViewer *viewer,
+                       GtkWidget *view,
+                       const gchar *title,
+                       gboolean needs_scrolled_window)
```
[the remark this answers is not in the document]
dnohales: This code was taken from `gis-endless-eula-page`, it's out of scope of this PR.

gnome-initial-setup/pages/endless-eula/gis-endless-eula-viewer.c
```diff
@@ -0,0 +1,356 @@
...
+                                                         NULL,
+                                                         &error);
+  if (error != NULL)
+    {
+      path = g_file_get_path (file);
+      g_critical ("Unable to load EvDocument for file %s: %s", path, error->message);
+      g_error_free (error);
+      g_free (path);
+
+      return NULL;
```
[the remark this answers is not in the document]
dnohales: This code was taken from gis-endless-eula-page, it's out of scope of this PR, but yeah, in this case it could be problematic.

gnome-initial-setup/pages/endless-eula/gis-endless-eula-viewer.c
dnohales: I've just taken the @ramcq's suggestion on file naming. In my opinion, the fact that is problematic to deal with strings in GLib in C even for simple concatenation should not interfiere with a file structure in another component.

If the structure you mention is better, I think it should be better for another reason. For me it's the same.

dnohales:
@wjt Updated.

gnome-initial-setup/pages/endless-eula/gis-endless-eula-viewer.c
wjt: Translations live in (eg) `/usr/share/$locale/LC_MESSAGES/gnome-initial-setup.mo`; the Help Center documents live in (eg) `/usr/share/help/$locale/gnome-help/index.page`. The pattern of using a single path component for the locale and leaving all others the same is widely used.

dnohales: Yes, I know. I think that's a good reason.

dnohales:
@wjt Updated this and the other PRs.

PR #272, opened 2018-11-28
T24470

cosimoc:
https://phabricator.endlessm.com/T24470


gnome-initial-setup/gis-driver.h
```diff
@@ -107,6 +107,8 @@ gboolean gis_driver_is_small_screen (GisDriver *driver);
 
 gboolean gis_driver_is_product (GisDriver *driver, const gchar *product_name);
 
+gboolean gis_driver_is_hack (GisDriver *driver);
```
wjt: Good catch.

PR #273, opened 2018-12-11
driver: Connect to GDM as soon as g-i-s starts

jprvita:
GDM can't tell Plymouth to quit until g-i-s connects to it, so GDM knows
g-i-s has started successfully. Currently g-i-s only connects to GDM on
the last page, when it logs-in the new user. This prevents accessing the
TTY and makes systemd account the time spent in g-i-s as part of the
boot process.

Connecting to GDM as soon as g-i-s allows GDM to tell Plymouth to quit
at the right time, avoiding the problems described above.

Closes: https://gitlab.gnome.org/GNOME/gdm/issues/439

https://phabricator.endlessm.com/T21181

### DIFF
--- a/gnome-initial-setup/gis-driver.c
+++ b/gnome-initial-setup/gis-driver.c
@@ -25,6 +25,7 @@
 
 #include <stdlib.h>
 #include <locale.h>
+#include <gdm/gdm-client.h>
 
 #include "gis-assistant.h"
 #include "gis-page-util.h"
@@ -74,6 +75,10 @@ static GParamSpec *obj_props[PROP_LAST];
 struct _GisDriverPrivate {
   GtkWindow *main_window;
   GisAssistant *assistant;
+
+  GdmClient *client;
+  GdmGreeter *greeter;
+  GdmUserVerifier *user_verifier;
 
   ActUser *user_account;
   gchar *user_password;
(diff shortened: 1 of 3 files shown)


gnome-initial-setup/gis-driver.c
```diff
@@ -904,6 +938,25 @@ emit_startup_sound (GApplication * app, gpointer user_data)
...
 
+void
+connect_to_gdm (GisDriverPrivate *priv)
+{
+  g_autoptr(GError) error = NULL;
+
+  priv->client = gdm_client_new ();
+
+  priv->greeter = gdm_client_get_greeter_sync (priv->client, NULL, &error);
+  if (!error)
```
wjt:
```suggestion
  if (error == NULL)
```

to match surrounding code – I'll apply this locally.

gnome-initial-setup/gis-driver.c
```diff
@@ -904,6 +938,26 @@ emit_startup_sound (GApplication * app, gpointer user_data)
                      G_DBUS_CALL_FLAGS_NONE, -1, NULL, NULL, NULL);
 }
 
+static void
```
wjt: I made this `static`.

gnome-initial-setup/gis-driver.c
```diff
@@ -904,6 +938,26 @@ emit_startup_sound (GApplication * app, gpointer user_data)
...
+connect_to_gdm (GisDriver *driver)
```
wjt: I made this accept the `GisDriver` rather than the `Priv` structure: this is more conventional.

PR #275, opened 2018-12-12
summary: Launch HackUnlock asynchronously

dnohales:
https://phabricator.endlessm.com/T24631

### DIFF
--- a/gnome-initial-setup/pages/summary/gis-summary-page.c
+++ b/gnome-initial-setup/pages/summary/gis-summary-page.c
@@ -191,9 +191,23 @@ log_user_in (GisSummaryPage *page)
 }
 
 static void
+set_busy (GisSummaryPage *page,
+          gboolean        busy)
+{
+  GisSummaryPagePrivate *priv = gis_summary_page_get_instance_private (page);
+  GdkWindow *window = gtk_widget_get_window (GTK_WIDGET (page));
+  GdkDisplay *display = gdk_window_get_display (window);
+  GdkCursor *cursor = gdk_cursor_new_from_name (display, busy ? "wait" : "default");
+
+  gtk_widget_set_sensitive (priv->start_button, !busy);
+  gdk_window_set_cursor (window, cursor);
+}
+
+static void
 finish_fbe (GisSummaryPage *page)
 {
   gis_ensure_stamp_files ();
+  set_busy (page, FALSE);
 
   switch (gis_driver_get_mode (GIS_PAGE (page)->driver))
     {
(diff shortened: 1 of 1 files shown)


cosimoc:
It's not a bad idea in general to make the code asynchronous, but I believe the design intention for the spinner was simply to change the mouse cursor, not necessarily add a spinner next to the button, which would not require the async change... Do you have a screenshot of how this looks?

dnohales:
I'd prefer to have it async, just because I don't feel the window blocking looks ok nor that is ok technically. I agree we can change the spinner with the busy cursor, that probably looks better, the current situation of the `GtkSpinner` not spinning is not ideal:

captura de pantalla de 2018-12-12 19-15-03

dnohales:
Updated.

gnome-initial-setup/pages/summary/gis-summary-page.c
```diff
@@ -268,21 +281,15 @@ run_initial_contact_quest (GisSummaryPage *page)
...
-                          "Connect",
-                          g_variant_new ("(sss)",
-                                         "view.JSContext.globalParameters",
-                                         "notify",
-                                         "unlocked"),
-                          G_DBUS_CALL_FLAGS_NONE,
-                          G_MAXINT,
-                          NULL,
-                          &error);
+  g_dbus_proxy_call_finish (G_DBUS_PROXY (proxy), res, &error);
```
cosimoc: Don't we need to call `set_busy(FALSE)` at the end of this? Or does the cursor change back to normal as a consequence of this window getting hidden and the unlock app starting?

dnohales: Ah, I did it before but for some reason I removed it. But yeah, the cursor went back to normal without `set_busy (page, FALSE);`, probably because of the way I did it, the cursor was only set for the g-i-s window, not sure.

Updated it anyway, looks more correct to reset it.

cosimoc: Cool! I don't mean to nitpick but we should probably also reset the cursor in the error code paths then.

dnohales: I added the `set_busy (page, FALSE);` line in `finish_fbe`, which is called on both the success and error path.

dnohales: I think, maybe I'm missing something.

cosimoc: My bad!

PR #280, opened 2019-02-18
Add product_serial file to the tmpfile.d

danigm:
This will allow us to use the product_serial for the account creation.

https://phabricator.endlessm.com/T25480

### DIFF
--- a/data/product_serial.conf
+++ b/data/product_serial.conf
@@ -1,1 +1,2 @@
 z    /sys/devices/virtual/dmi/id/product_uuid 0444 root nogroup - -
+z    /sys/devices/virtual/dmi/id/product_serial 0444 root nogroup - -


danigm:
I've just updated the commit message with the one you proposed. We don't have phabricator ticket yet.

wjt:
Please open one.

danigm:
I've created the task and added to the commit message.

wjt:
@dsd as someone who was involved in the initial work to make product_serial, then product_uuid, readable initially by nogroup then by all users (see tickets linked on 3513822675d9ac64d40ffb3d959594b4c9e88b6e) do you have any opinion about the implications of making the product serial number world-readable? I guess the main concern is a privacy one, but I don't think the product serial number is substantially more sensitive than the (equally unique) product UUID.

dsd:
I don't view the reasoning of "its fine because we do it for uuid" as very strong - when the product was at a much earlier stage years ago, we would often sidestep concerns like this one in the name of getting a product to market as quick as possible. Glancing at the old issues it doesn't look like the privacy issue was really considered or discussed.

That said though, I don't have specific concerns with making this information accessible.

wjt:
Sure – FWIW I also think it's not an amazing argument, even though I'm employing it.

There is a similar time pressure around this feature so I'm going to merge this change, but will follow up on the Phabricator ticket about the privacy implications of this and the existing product_uuid change.

PR #281, opened 2019-03-05
T25318 greet user on desktop

joaquimrocha:
https://phabricator.endlessm.com/T25318

### DIFF
--- a/gnome-initial-setup/gnome-initial-setup.c
+++ b/gnome-initial-setup/gnome-initial-setup.c
@@ -318,14 +318,6 @@ main (int argc, char *argv[])
     return EXIT_SUCCESS;
   }
 
-  /* Upstream has "existing user" mode for new user accounts, but we don't
-     want that in Endless, so skip it. In the future, we should be launching
-     the tutorial right from here, once it's again available. */
-  if (get_mode () == GIS_DRIVER_MODE_EXISTING_USER) {
-    gis_ensure_stamp_files ();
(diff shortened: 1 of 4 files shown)


gnome-initial-setup/gnome-initial-setup.c
```diff
@@ -345,6 +337,23 @@ main (int argc, char *argv[])
...
+     want that in Endless, so skip it. In the future, we should be launching
+     the tutorial right from here, once it's again available. */
+  if (get_mode () == GIS_DRIVER_MODE_EXISTING_USER) {
+    if (gis_driver_is_hack (driver)) {
+      g_autoptr(GError) error = NULL;
+      if (!g_application_register (G_APPLICATION (driver), NULL, &error))
+        g_warning("Cannot run Welcome quest due to failed g-i-s register: %s",
+                  error->message);
+      else
+        gis_driver_run_clubhouse_quest (driver, "AdaQuestSet.Welcome");
```
cosimoc: The other two cleanup commits may still be worth picking, but as commented on the ticket I don't think that this is the right place to accomplish what you need.

joaquimrocha: I don't think the other two commits are needed if we're not adding the 3rd. But I will let @wjt decide. I am fine closing this PR and moving the logic for the Welcome quest launcher into the Shell as @cosimoc advised.

PR #282, opened 2019-04-25
Unlock keyring using libsecret + internal D-Bus API

wjt:
libgnome-keyring has been dropped from Debian in favour of applications
using libsecret.
<https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=892359>

libsecret does not provide a straightforward API to unlock a keyring
with a given passphrase, but we can do it with gnome-keyring's
InternalUnsupportedGuiltRiddenInterface, which gis-keyring.c uses to set
the passphrase on the login keyring in the first place.

This should be squashed into 'copy-worker: Unlock keyring on first
login' in the next rebase.

https://phabricator.endlessm.com/T26274

### DIFF
--- a/gnome-initial-setup/gnome-initial-setup-copy-worker.c
+++ b/gnome-initial-setup/gnome-initial-setup-copy-worker.c
@@ -6,7 +6,6 @@
 #include <pwd.h>
 #include <string.h>
 #include <gio/gio.h>
-#include <gnome-keyring.h>
 #include <stdlib.h>
 
 static char *
@@ -66,23 +65,6 @@ move_file_from_homedir (GFile       *src_base,
   }
 }
 
-static void
-unlock_keyring (const gchar *initial_setup_homedir)
-{
-  gchar *file;
-  gchar *password = NULL;
-
-  file = g_build_filename (initial_setup_homedir, ".config/password", NULL);
-  if (g_file_get_contents (file, &password, NULL, NULL))
-    gnome_keyring_unlock_sync ("login", password);
-  else
-    g_warning ("Unable to read user password file");
-
-  g_remove (file);
-  g_free (file);
-  g_free (password);
-}
-
 int
 main (int    argc,
       char **argv)
@@ -113,8 +95,6 @@ main (int    argc,
   FILE (".config/goa-1.0/accounts.conf");
   FILE (".config/monitors.xml");
   FILE (".local/share/keyrings/login.keyring");
-
-  unlock_keyring (initial_setup_homedir);
 
   return EXIT_SUCCESS;
 }
--- a/gnome-initial-setup/meson.build
+++ b/gnome-initial-setup/meson.build
@@ -49,7 +49,6 @@ dependencies = [
     dependency ('gio-unix-2.0', version: '>= 2.53.0'),
     dependency ('gdm', version: '>= 3.8.3'),
     dependency ('geocode-glib-1.0'),
-    dependency ('gnome-keyring-1'),
     dependency ('libgeoclue-2.0', version: '>= 2.3.1'),
     cc.find_library('m', required: false),
     dependency ('pango', version: '>= 1.32.5'),


gnome-initial-setup/gnome-initial-setup-copy-worker.c
```diff
@@ -69,18 +71,47 @@ move_file_from_homedir (GFile       *src_base,
 static void
 unlock_keyring (const gchar *initial_setup_homedir)
 {
-  gchar *file;
-  gchar *password = NULL;
+  g_autofree gchar *file = g_build_filename (initial_setup_homedir, ".config/password", NULL);
```
pwithnall:
```suggestion
  g_autofree gchar *file = g_build_filename (initial_setup_homedir, ".config", "password", NULL);
```

Though I appreciate this is not your bug.

gnome-initial-setup/gnome-initial-setup-copy-worker.c
```diff
@@ -69,18 +71,47 @@ move_file_from_homedir (GFile       *src_base,
...
+    g_warning ("Unable to read user password file: %s", error->message);
+    goto out;
+  }
 
-  file = g_build_filename (initial_setup_homedir, ".config/password", NULL);
-  if (g_file_get_contents (file, &password, NULL, NULL))
-    gnome_keyring_unlock_sync ("login", password);
-  else
-    g_warning ("Unable to read user password file");
+  service = secret_service_get_sync (SECRET_SERVICE_OPEN_SESSION, NULL, &error);
```
pwithnall: Is it OK to make blocking D-Bus calls here? Is this function going to block the UI?

(I have no idea, but thought I should raise it anyway.)

pwithnall: I realise that if this is a problem, the refactoring will be non-trivial and can probably be deferred.

gnome-initial-setup/gnome-initial-setup-copy-worker.c
```diff
@@ -69,18 +71,47 @@ move_file_from_homedir (GFile       *src_base,
...
+  if (service == NULL) {
+    g_warning ("Failed to get secret service: %s", error->message);
+    goto out;
+  }
+
+  secret = secret_value_new (password, password_size, "text/plain");
+
+  g_dbus_connection_call_sync (g_dbus_proxy_get_connection (G_DBUS_PROXY (service)),
+                               "org.gnome.keyring",
```
pwithnall: Presumably you’re using the well-known name `org.gnome.keyring` rather than `org.freedesktop.secrets`, because you can’t guarantee that `org.gnome.keyring.InternalUnsupportedGuiltRiddenInterface` is not guaranteed to be exposed on `org.freedesktop.secrets` (if it was implemented by the KDE daemon, for example)?

Might be worth a comment to that effect.

gnome-initial-setup/gnome-initial-setup-copy-worker.c
```diff
@@ -69,18 +71,47 @@ move_file_from_homedir (GFile       *src_base,
...
+                               "/org/freedesktop/secrets",
+                               "org.gnome.keyring.InternalUnsupportedGuiltRiddenInterface",
+                               "UnlockWithMasterPassword",
+                               g_variant_new ("(o@(oayays))",
+                                              "/org/freedesktop/secrets/collection/login",
+                                              secret_service_encode_dbus_secret (service, secret)),
+                               NULL,
+                               0,
```
pwithnall:
```suggestion
                               G_DBUS_CALL_FLAGS_NONE,
```

gnome-initial-setup/gnome-initial-setup-copy-worker.c
```diff
@@ -69,18 +71,47 @@ move_file_from_homedir (GFile       *src_base,
...
+                               G_MAXINT,
```
pwithnall: No timeout? Brave.

gnome-initial-setup/gnome-initial-setup-copy-worker.c
```diff
@@ -69,18 +71,47 @@ move_file_from_homedir (GFile       *src_base,
...
-  gchar *password = NULL;
+  g_autofree gchar *file = g_build_filename (initial_setup_homedir, ".config/password", NULL);
+  gsize password_size;
+  g_autofree gchar *password = NULL;
+  SecretValue *secret = NULL;
+  SecretService *service = NULL;
+  g_autoptr(GError) error = NULL;
+
+  if (!g_file_get_contents (file, &password, &password_size, &error)) {
+    g_warning ("Unable to read user password file: %s", error->message);
```
pwithnall: (Existing bug, but:) Should we also check whether `password_size > 0`, and error otherwise?

gnome-initial-setup/gnome-initial-setup-copy-worker.c
pwithnall: Might also want to add a runtime dependency on `gnome-keyring` in `debian/control` too, for completeness.

gnome-initial-setup/gnome-initial-setup-copy-worker.c
```diff
@@ -2,12 +2,14 @@
 
 /* Copies settings installed from gnome-initial-setup and
  * sticks them in the user's profile */
+#include "config.h"
 
 #include <pwd.h>
 #include <string.h>
+#include <glib/gstdio.h>
 #include <gio/gio.h>
-#include <gnome-keyring.h>
```
pwithnall: From a quick `git grep`: `gnome-initial-setup/meson.build` appears to have a `dependency ('gnome-keyring-1')` which might need to be dropped?

gnome-initial-setup/gnome-initial-setup-copy-worker.c
wjt: No. This is a stand-alone program that runs (conditionally) at login. See `/usr/etc/xdg/autostart/gnome-initial-setup-copy-worker.desktop`.

There are reasons to feel sad about the way this works but I don't think this particular point is a problem. (Don't worry, though: the UI also makes blocking D-Bus calls constantly.)

gnome-initial-setup/gnome-initial-setup-copy-worker.c
wjt: No, it is legitimate for a user to have an empty passphrase.

But… haaang on! I can't see where this file is *created*.

gnome-initial-setup/gnome-initial-setup-copy-worker.c
wjt: For the record I'm using the same well-known name as the existing upstream code in `gis-keyring.c` does; yes, presumably that reason.

Dependency – I suppose for completeness one should do that but I don't think it's worth a downstream patch. I'll send a patch to Debian.

wjt: https://salsa.debian.org/gnome-team/gnome-initial-setup/merge_requests/4

gnome-initial-setup/gnome-initial-setup-copy-worker.c
wjt: OK – the file is not created, so all this code was unused, and yet the keyring is still unlocked. I have explained why in a new commit message which deletes this code instead.

wjt:
Updated with a new "implementation" which just removes this code.

In the version of this patch on the eos3.1 branch and earlier branches, 5284f2078a906170fafc78ad09889a7a5e9e38f9, the account page wrote the password to disk. The eos3.2 rebase brought in an upstream change where the account page was split into "local" and "enterprise" bits; the version of this patch on that branch, 38ca1bdac31fbddaf095efcb84c0e3ff854b04cd, does not include the hunk to write the password file. So since July 2017 this code has been a no-op.

I think this implementation is pretty dodgy, anyway – I think these files should be copied into place before the session starts, to avoid the faintest hint of a situation where gnome-keyring auto-creates the login keyring, then the copy-worker overwrites it. This should be an upstream change for all the other files too.

PR #284, opened 2019-09-03
gnome-initial-setup rebase, Debian packaging

ptomato:
https://phabricator.endlessm.com/T25162


debian/control
```diff
@@ -10,7 +10,6 @@ Uploaders: Jeremy Bicha <jbicha@debian.org>
 Build-Depends: debhelper-compat (= 12),
                dh-sequence-gnome,
                geoclue-2.0 (>= 2.3.1),
-               iso-codes,
```
wjt: I think this should be submitted to the Debian packaging.

(On the other hand, I am pretty sure that, at runtime, gnome-initial-setup relies on iso-codes being installed because libgnome-desktop's `gnome-languages.c` reads `/usr/share/xml/iso-codes/iso_3166.xml`; weirdly, the `libgnome-desktop-3-17` package doesn't have this dependency.)

debian/rules
```diff
@@ -8,6 +8,7 @@ export DEB_LDFLAGS_MAINT_APPEND = -Wl,-z,defs -Wl,--as-needed
 
 override_dh_auto_configure:
 	dh_auto_configure -- \
+		--libexecdir=/usr/lib/gnome-initial-setup \
```
wjt: I could take this or leave it… I'm reasonably sure nothing outside gnome-initial-setup relies on this path and it's one more bit of divergence to carry forever, but OTOH it's not a big patch to carry.

debian/control
ptomato: This seems to have been added in https://salsa.debian.org/gnome-team/gnome-initial-setup/commit/85732c17ee37e93b568a933cac5589ce853d596c at a time when GDM was a build dependency, as far as I can tell from the commit message?

debian/rules
ptomato: OK, the only thing I can find inside gnome-initial-setup is [this](https://github.com/endlessm/gnome-initial-setup/blob/master/data/20-gnome-initial-setup.rules#L27), which should be deleted anyway as that script doesn't seem to exist. I'm a bit concerned that changing this might invalidate all the testing I've done so far, since it seems like the likely failure points would be in the "unusual" modes such as demo mode...

wjt: That script was a precursor to the `eos-transient-setup.service` systemd unit. Indeed it should go away.

OTOH, https://github.com/endlessm/eos-event-recorder-daemon/pull/234 would be such a place outside gnome-initial-setup :-)

debian/control
ptomato: Submitted upstream: https://salsa.debian.org/gnome-team/gnome-initial-setup/merge_requests/8

PR #286, opened 2019-09-18
account: Use control-center setting to override faces

danigm:
This patch uses the gnome-control-center user-accounts facesdir
configuration to override the default faces dir.

https://phabricator.endlessm.com/T27788

### DIFF
--- a/gnome-initial-setup/pages/account/um-photo-dialog.c
+++ b/gnome-initial-setup/pages/account/um-photo-dialog.c
@@ -42,9 +42,6 @@
 #define ROW_SPAN 5
 #define AVATAR_PIXEL_SIZE 72
 
-#define CONFIG_ACCOUNT_GROUP "page.account"
-#define CONFIG_ACCOUNT_FACESDIR_KEY "faces"
-
 struct _UmPhotoDialog {
         GtkPopover parent;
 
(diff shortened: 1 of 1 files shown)


gnome-initial-setup/pages/account/um-photo-dialog.c
```diff
@@ -203,6 +203,10 @@ get_facesdirs (void)
         const char *facesdir_env;
         GisDriver *driver;
 
+        GSettingsSchemaSource *source = g_settings_schema_source_get_default ();
+        g_autoptr(GSettingsSchema) schema = NULL;
+        g_autoptr(GSettings) settings = NULL;
```
ptomato: Small nitpick, I would move this inside the `if (schema)` block so that it is freed earlier when the block ends.

danigm:
I've updated the code with the same code from the `gnome-control-center` to allow more than one faces directories.

gnome-initial-setup/pages/account/um-photo-dialog.c
ptomato: This comment no longer makes sense in the updated code.

gnome-initial-setup/pages/account/um-photo-dialog.c
```diff
@@ -195,36 +192,48 @@ create_face_widget (gpointer item,
...
-                                             g_strdup (facesdir_env));
+        GSettingsSchemaSource *source = g_settings_schema_source_get_default ();
+
+        if (!source)
+                return facesdirs;
+
+        schema = g_settings_schema_source_lookup (source,
+                                                  "org.gnome.ControlCenter",
+                                                  FALSE);
+        if (!schema)
```
ptomato: I think this part should change now, I'm pretty sure you can assume that the `org.gnome.ControlCenter` schema is present, but you should check for the key and return `facesdirs` in that case so that you don't abort if the FBE is used with a mismatched version of Settings.

danigm: Done

gnome-initial-setup/pages/account/um-photo-dialog.c
```diff
@@ -195,36 +192,58 @@ create_face_widget (gpointer item,
...
+
+        schema = g_settings_schema_source_lookup (source,
+                                                  "org.gnome.ControlCenter",
+                                                  FALSE);
 
-        facesdir_env = g_getenv ("GIS_ACCOUNT_PAGE_FACESDIR");
-        if (facesdir_env != NULL && g_strcmp0 (facesdir_env, "") != 0) {
-                facesdirs = g_slist_prepend (facesdirs,
-                                             g_strdup (facesdir_env));
+        keys = g_settings_schema_list_keys (schema);
```
ptomato: [`g_settings_schema_has_key()`](https://developer.gnome.org/gio/stable/gio-GSettingsSchema-GSettingsSchemaSource.html#g-settings-schema-has-key)?

danigm: Yes, sure that's better :sweat_smile: I was looking for that in the doc, but I didn't find it because I was looking here: https://developer.gnome.org/gio/stable/GSettings.html

PR #294, opened 2019-10-22
summary: update check mark icon for new default background

erikos:
https://phabricator.endlessm.com/T28188


wjt:
Have you verified how this looks in the app? I ask because when I open the SVGs in Firefox, the old one correctly displays as 256×256, whereas this one is gigantic.

erikos:
Yeah, dropbox displays the file correctly and I can view it in Inkscape correctly too. It says it is 256x256. In the diff it shows correct too.

wjt:
Dropbox, Inkscape and the GitHub diff are not the app itself. It does not look correct in the app.

Screenshot from 2019-10-22 12-17-35

erikos:
Sure, just saying that I tested that so far and that I have seen some differences in the visual output there. I guess different interpretation of the svg.

wjt:
Well – I spent some time tinkering in Inkscape and I got something that renders properly in the app.

https://willthompson.co.uk/misc/ready-to-go.svg

But I think it looks really bad:

Screenshot from 2019-10-23 22-38-46

Compare it to what we have now:

Screenshot from 2019-10-23 22-42-11

On closer inspection, the check mark is actually not the Endless wallpaper – it's an upstream asset. The new Endless wallpaper would be shown behind the app anyway so… is what we actually want to make that ✔ part of the window transparent?

Anyway ­– we could spend time tinkering with this, but personally I think the time would be better spent pursuing something like what's proposed at https://gitlab.gnome.org/GNOME/gnome-initial-setup/issues/8 and keeping the upstream asset for now.

erikos:
Interesting, I always thought that asset was there from an old Endless default background. 

Agreed let's invest in the upstream change.

PR #297, opened 2019-12-20
Fix "Password hint" translation to pt_BR

andrunko:
Also needs to be cherry-picked to 3.7 which has the same issue.

https://phabricator.endlessm.com/T29000


andrunko:
This has to be done in transifex instead, closing here.

PR #299, opened 2020-02-10
network: react to devices being connected

wjt:
A clean backport of
https://gitlab.gnome.org/GNOME/gnome-initial-setup/merge_requests/69,
complicated only by my own downstream conflicting change. This branch
reverts that change, backports !69, and then applies a refreshed version
of that patch which I have posted to
https://gitlab.gnome.org/GNOME/gnome-initial-setup/merge_requests/74.

https://phabricator.endlessm.com/T27527

### DIFF
--- a/gnome-initial-setup/pages/network/gis-network-page.c
+++ b/gnome-initial-setup/pages/network/gis-network-page.c
@@ -56,6 +56,9 @@ struct _GisNetworkPagePrivate {
   GtkSizeGroup *icons;
 
   guint refresh_timeout_id;
+
+  /* TRUE if the page has ever been shown to the user. */
+  gboolean ever_shown;
 };
 typedef struct _GisNetworkPagePrivate GisNetworkPagePrivate;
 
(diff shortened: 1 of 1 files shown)


gnome-initial-setup/pages/network/gis-network-page.c
```diff
@@ -601,17 +611,62 @@ active_connections_changed (NMClient *client, GParamSpec *pspec, GisNetworkPage
...
+  if (priv->nm_device == NULL) {
+    has_device = FALSE;
+    activated = FALSE;
+    connectivity_check_complete = FALSE;
+  } else {
+    has_device = TRUE;
+    device_state = nm_device_get_state (priv->nm_device);
+    connectivity = nm_client_get_connectivity (priv->nm_client);
+    activated = device_state == NM_DEVICE_STATE_ACTIVATED;
+    connectivity_check_complete = connectivity != NM_CONNECTIVITY_NONE;
```
pwithnall: Should it not be checking against `NM_CONNECTIVITY_UNKNOWN`? The NM API documentation says that `UNKNOWN` is used when the connectivity check hasn’t run yet.

gnome-initial-setup/pages/network/gis-network-page.c
```diff
@@ -59,6 +59,16 @@ struct _GisNetworkPagePrivate {
 
   /* TRUE if the page has ever been shown to the user. */
   gboolean ever_shown;
+
+  /* If nm_device is in state ACTIVATED, but nm_client's connectivity is not
+   * yet FULL, the event source ID for a timeout. 0 otherwise, or after the
+   * timeout fires.
+   */
+  guint wait_connected_timeout_id;
```
pwithnall: This needs to be removed and cleared in `dispose()` too, just in case the page somehow gets destroyed before it fires.

wjt: Resolved in upstream version.

gnome-initial-setup/pages/network/gis-network-page.c
wjt: https://gitlab.gnome.org/GNOME/gnome-initial-setup/merge_requests/74#note_711446 – I'm going to drop this patch downstream for now. In my opinion, it is more important that we notice new devices than to preserve this attempt to make the timezone page show up more rarely, pending more thought on how to make this work well.

pwithnall: :+1:

pwithnall:
I see you’ve dropped the refreshed backport of upstream’s !74 from this branch. However, I think the commit message for “Revert "network: wait 3 seconds for connectivity check” should be tweaked to mention that the refreshed backport is not going to be applied, and the 3s wait is being dropped indefinitely because it’s hard to get working.

LGTM to merge after that commit message is tweaked (no need for rereview as far as I’m concerned).

PR #302, opened 2020-03-13
fixup! account-local: add toggle for passwordless accounts

mwleeds:
Normally the password page is hidden when the "Set a password" checkbox
is left unchecked on the Account page. In
gis_account_page_local_apply(), we call gis_driver_set_username() and
gis_driver_set_passwordless() and these both trigger the
username_or_passwordless_changed() callback in the password page.
However, when the parental controls checkbox on the account page is
checked, gis_driver_set_parental_controls_enabled() is called which
leads to all the pages being destroyed and rebuilt (see
rebuild_pages()), which means we lose the "hidden" state of the password
page when it should be hidden. So call
username_or_passwordless_changed() during construction of the password
page object as well, and use g_signal_connect_swapped() so we can avoid
faking the pspec parameter.

https://phabricator.endlessm.com/T28739

### DIFF
--- a/gnome-initial-setup/pages/password/gis-password-page.c
+++ b/gnome-initial-setup/pages/password/gis-password-page.c
@@ -384,13 +384,11 @@ confirm_changed (GtkWidget      *w,
 }
 
 static void
-username_or_passwordless_changed (GObject         *obj,
-                                  GParamSpec      *pspec,
-                                  GisPasswordPage *page)
+username_or_passwordless_changed (GisPasswordPage *page)
 {
   GisPasswordPagePrivate *priv = gis_password_page_get_instance_private (page);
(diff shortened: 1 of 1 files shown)


gnome-initial-setup/pages/password/gis-password-page.c
```diff
@@ -477,7 +475,8 @@ gis_password_page_constructed (GObject *object)
                                                               "org.freedesktop.DBus.Properties",
                                                               NULL, NULL);
 
-  gtk_widget_show (GTK_WIDGET (page));
+  /* This callback shows the page widget conditionally on if it's needed */
```
pwithnall: Nitpick: ‘only’?

mwleeds: I think "conditional on" is a legitimate phrase

PR #303, opened 2020-03-16
eos: Report metrics from parental controls page

pwithnall:
If the user chooses to enable parental controls when setting up the
computer, send a metrics event which contains the initial parental
controls settings. Don’t send an event if parental controls are not
enabled.

This will give us an idea of the more and less popular parental controls
features. If they are compared against parental controls settings set or
changed after initial setup, this metric will also give us an idea of
how people change their parental controls (presumably tweaking them
since they set them incorrectly in the first place).

This commit should not be upstreamed.

Signed-off-by: Philip Withnall <withnall@endlessm.com>

https://phabricator.endlessm.com/T28742

### DIFF
--- a/gnome-initial-setup/pages/parental-controls/gis-parental-controls-page.c
+++ b/gnome-initial-setup/pages/parental-controls/gis-parental-controls-page.c
@@ -25,6 +25,7 @@
 
 #include "config.h"
 
+#include <eosmetrics/eosmetrics.h>
 #include <glib.h>
 #include <glib-object.h>
 #include <glib/gi18n.h>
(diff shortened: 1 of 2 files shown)


pwithnall:
Depends on https://gitlab.freedesktop.org/pwithnall/malcontent/-/merge_requests/45 (and a backport of that to EOS). Currently untested other than to check that it compiles.

gnome-initial-setup/pages/parental-controls/gis-parental-controls-page.c
```diff
@@ -46,6 +47,24 @@ struct _GisParentalControlsPage
 
 G_DEFINE_TYPE (GisParentalControlsPage, gis_parental_controls_page, GIS_TYPE_PAGE)
 
+#define GIS_PARENTAL_CONTROLS_METRIC "5eae336c-f431-43f3-8e17-e0976b856bd8"
```
wjt: I did have a passing thought about how we will distinguish this event from the same event submitted by something in the user session. Actually this can be done because events include the Unix UID that submitted them.

One purely bikeshed based comment based on the convention used in other projects:

```suggestion
#define GIS_PARENTAL_CONTROLS_EVENT "5eae336c-f431-43f3-8e17-e0976b856bd8"
```

pwithnall: :+1:

pwithnall:
Updated to bump the malcontent dependency to 0.7.0 for the new API: Jenkins should pass now. Debian changes in #308.

mwleeds:
bot, retest this please

pwithnall:
Sigh, #309 is needed too, sorry.

mwleeds:
bot, retest this please

PR #305, opened 2020-03-16
Fix crashes when saving parental controls settings

mwleeds:
These are cherry picks from this upstream MR: https://gitlab.gnome.org/GNOME/gnome-initial-setup/-/merge_requests/80

https://phabricator.endlessm.com/T28787


pwithnall:
I think this can be merged as soon as it’s accepted upstream.

pwithnall:
[Merged upstream](https://gitlab.gnome.org/GNOME/gnome-initial-setup/-/merge_requests/80#note_741951).

PR #306, opened 2020-03-18
fixup! password: Add eye icons to toggle password visibility

mwleeds:
Make the "show password" icon match the one in GNOME Shell, rather than
being reversed.

https://phabricator.endlessm.com/T29262

### DIFF
--- a/gnome-initial-setup/pages/password/gis-password-page.c
+++ b/gnome-initial-setup/pages/password/gis-password-page.c
@@ -327,11 +327,15 @@ on_entry_icon_press (GtkEntry            *entry,
   if (icon_pos != GTK_ENTRY_ICON_SECONDARY)
     return;
 
+  /* We show the eye icon with a slash through it when the password is not
+   * visible, which is consistent with how gnome-shell works in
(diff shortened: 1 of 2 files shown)


gnome-initial-setup/pages/password/gis-password-page.c
```diff
@@ -327,21 +327,25 @@ on_entry_icon_press (GtkEntry            *entry,
   if (icon_pos != GTK_ENTRY_ICON_SECONDARY)
     return;
 
+  /* We show the eye icon with a slash through it when the password is not
+   * visible, which is consistent with how gnome-shell works in
+   * StPasswordEntry.
+   */
   if (gtk_entry_get_visibility (entry))
```
andrunko: We need to reverse the check here also no?

mwleeds: No, then we'd be back to where we started. A double negative is a positive

mwleeds: Never mind, you're right actually

PR #312, opened 2020-04-06
fixup! eos: Report metrics from parental controls page

pwithnall:
This should be squashed in when we next rebase.

It reworks the metrics submission for parental controls to use the same
event (data format and UUID) as the event submitted by
malcontent-control when a user’s parental controls are changed. A
`IsInitialSetup` field differentiates the two.

This allows shared code for processing the events on the server, and
allows some more comparisons to be made between initial-setup and
malcontent-control events.

https://phabricator.endlessm.com/T28742


pwithnall:
Currently untested due to https://phabricator.endlessm.com/T29711, but open for review anyway.

PR #313, opened 2020-04-13
account: Use custom setting to override faces

danigm:
This patch uses the org.gnome.desktop.interface.facesdirs
configuration to override the default faces dir.

https://phabricator.endlessm.com/T29718

### DIFF
--- a/gnome-initial-setup/pages/account/um-photo-dialog.c
+++ b/gnome-initial-setup/pages/account/um-photo-dialog.c
@@ -217,72 +217,72 @@ create_face_widget (gpointer item,
         return image;
 }
 
-static GSList *
+static GStrv
 get_settings_facesdirs (void)
 {
-        g_autoptr(GSettings) settings = NULL;
         g_autoptr(GSettingsSchema) schema = NULL;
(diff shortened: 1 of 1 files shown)


wjt:
We already have 256b0c09cdbfd81fdd5b59915a8d1e044bf83ead on this branch – it looks like the patch in this PR updates the implementation to match what's on https://gitlab.gnome.org/GNOME/gnome-initial-setup/-/merge_requests/60.

I left some feedback on that upstream MR. I don't think this change is necessary because the existing patch works:

Capture d’écran de 2020-04-15 15-57-22

andrunko:
> Exact same comments as the ones from [endlessm/gnome-control-center#217 (review)](https://github.com/endlessm/gnome-control-center/pull/217#pullrequestreview-393868144).
> 
> Specially about the potential issue when calling `g_list_store_append (faces, face_file);` from add_faces_from_dirs.

Ignore the above, comments clarified in the upstream PR.

danigm:
Okay, so we don't need this PR, I'll close the PR.

PR #315, opened 2020-04-29
fixup! account: Use control-center setting to override faces

GeorgesStavracas:
The upstream setting name is 'avatar-directories' now

https://phabricator.endlessm.com/T29718


andrunko:
LGTM.

PR #317, opened 2020-08-17
T30173 Rebase on 3.37.92

pwithnall:
This does most of the work for https://phabricator.endlessm.com/T30173, but isn’t actually the final rebase on 3.38.0 because that hasn’t been released yet.

Notable things to point out: demo mode has been dropped (as per https://phabricator.endlessm.com/T30173#840626), and the barcode has been dropped from the factory dialogue (as per https://phabricator.endlessm.com/T20124). Various smaller patches have been submitted upstream.

Packaging changes are in #318.

Here’s what happened to all the commits (oldest first), as a result of `git rebase -i upstream/3.37.91`:
 * **drop** `9378ab08 Allow to reorder pages from the vendor configuration file` — never used as per https://gitlab.gnome.org/GNOME/gnome-initial-setup/-/merge_requests/48#note_677130
 * *pick* `a0576771 copypasta code: Use icon name instead of stock icon` — keep, can’t currently be upstreamed due to being blocked on https://gitlab.gnome.org/GNOME/gnome-initial-setup/-/issues/68
 * *pick* `ab42158f copypasta code: Use GTask in UmRealmManager` — keep, can’t currently be upstreamed due to being blocked on https://gitlab.gnome.org/GNOME/gnome-initial-setup/-/issues/68
 * *pick* `7f2bc934 Allow initial setup to switch metrics on and off` — keep, Endless-specific
 * **drop** `2fab2b48 summary: do not play welcome tour` — dropped in favour of https://gitlab.gnome.org/GNOME/gnome-initial-setup/-/merge_requests/92, plus addition of `run_welcome_tour=false` to `debian/vendor.conf`
 * *pick* `615a973f Add product_serial.conf file` — keep, Endless-specific
 * *pick* `598105db account-local: limit full name to 80 characters` — keep, Endless-specific, and I don’t have the time to work through https://phabricator.endlessm.com/T25444 right now
 * *pick* `1ab4607e account-local: Hide username` — keep, Endless-specific for now
 * *pick* `e1d80c6d account-local: create a shared user account` — keep, Endless-specific for now
 * *pick* `23aff0a8 um-utils: consider "shared" username to be in use` — keep, Endless-specific for now
 * *pick* `40fa4f60 main,summary: skip initial setup entirely for shared account` — keep, Endless-specific for now
 * *pick* `c3915a4f account-local: username suggestions` — keep, but I have no idea what this is for and have not found time to investigate
[...]

PR #318, opened 2020-08-17
T30173 Rebase on 3.37.92 (packaging)

pwithnall:
This does most of the work for https://phabricator.endlessm.com/T30173, but isn’t actually the final rebase on 3.38.0 because that hasn’t been released yet.

This is the result of `git rebase -i salsa/debian/master`, which currently points to the [upload of the new `3.37.91-1` release](https://salsa.debian.org/gnome-team/gnome-initial-setup/-/commit/a8c20c4d1a22dc5d6c6a59730f786549b4136da0).

The barcode from the factory dialogue has been dropped (as per https://phabricator.endlessm.com/T20124) so the `zint` dependency is gone. I haven’t applied `debian/patches/revert_gnome_tour.patch` because we disable the tour, so using Yelp instead of gnome-tour is irrelevant. Other than that there’s not much to report.

Code changes are in #317.

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,3 +1,10 @@
+gnome-initial-setup (3.37.92-1endless0) eos; urgency=medium
+
+  * Endless 3.37.92 release
+    - Rebase on 3.37.92 upstream release
+
+ -- Philip Withnall <withnall@endlessm.com>  Tue, 08 Sep 2020 11:47:00 +0100
+
 gnome-initial-setup (3.37.91-1endless0) eos; urgency=medium
 
   * Endless 3.37.91 release


pwithnall:
bot, retest this please

GeorgesStavracas:
It looks like Flatpak needs to be added somewhere (either as `libflatpak-dev` in the build dependencies, or as `libflatpak` in the runtime dependencies) since it's a requirement for malcontent.

pwithnall:
> It looks like Flatpak needs to be added somewhere (either as `libflatpak-dev` in the build dependencies, or as `libflatpak` in the runtime dependencies) since it's a requirement for malcontent.

What makes you say that? The OBS build [succeeded](https://ci.endlessm-sf.com/job/obs-gnome-initial-setup-pr/243/). The `pkgconfig(flatpak)` dependency in `.gitlab-ci.yml` is only there because the upstream CI also builds malcontent (rather than installing it as a package), because the version g-i-s depends on isn’t in Fedora yet.

GeorgesStavracas:
It failed to build in my personal OBS branch, since malcontent depends on flatpak but it isn't installed.

pwithnall:
Do you have a build log?

Is it failing in the build, or in the unit tests?

pwithnall:
Pushed a trivial update to increment the version number since I updated #317 to rebase on 3.37.92 rather than 3.37.91.

mwleeds:
>  I haven’t applied debian/patches/revert_gnome_tour.patch because we disable the tour, so using Yelp instead of gnome-tour is irrelevant. Other than that there’s not much to report.

Shouldn't we drop it from `debian/patches/series` then? Since, as you say, we don't need it, and also since my understanding is we want all patches on the code branch and not the packaging branch, even if they come from Debian's packaging.

pwithnall:
> Shouldn't we drop it from `debian/patches/series` then? Since, as you say, we don't need it, and also since my understanding is we want all patches on the code branch and not the packaging branch, even if they come from Debian's packaging.

I think that would add extra maintenance overhead (divergence in `d/p/series`) for no benefit (our build system completely ignores `d/p/series`, so it’s not currently causing a problem).

PR #320, opened 2020-09-11
T30173 Misc use-after-free fixes

pwithnall:
These won’t fix T30173 but make debugging it a little less noisy.

All upstream in https://gitlab.gnome.org/GNOME/gnome-initial-setup/-/merge_requests/103 except the `UmRealmManager` fixes which fix a bug introduced in our downstream commit “copypasta code: Use GTask in UmRealmManager”.


pwithnall:
Merged upstream: https://gitlab.gnome.org/GNOME/gnome-initial-setup/-/merge_requests/103#note_c1c16af5b1aa0dfe9896ddc4ef6e83a36a0d361b

PR #321, opened 2020-09-11
debian: Disable Cheese webcam support to avoid a pipewire crash

pwithnall:
This can be re-enabled once the upstream pipewire bug is fixed:
https://gitlab.freedesktop.org/pipewire/pipewire/-/issues/286

Signed-off-by: Philip Withnall <withnall@endlessm.com>

https://phabricator.endlessm.com/T30173

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,3 +1,12 @@
+gnome-initial-setup (3.37.92-1endless2) eos; urgency=medium
+
+  * Disable Cheese webcam support to avoid a pipewire crash
+    This can be re-enabled once the upstream pipewire bug is fixed:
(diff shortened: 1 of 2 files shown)


pwithnall:
Actually, maybe don’t merge this: Wim has just [fixed the pipewire issue upstream](https://gitlab.freedesktop.org/pipewire/pipewire/-/commit/f6bc9113cef765d363e8af6e0e8d3379c8b09c1e), so I’ll try backporting that patch first.

pwithnall:
Yes, https://obs-master.endlessm-sf.com/request/show/21661 actually fixes the crash, so this workaround is not needed.

PR #324, opened 2020-10-13
Revert "account: Check configuration to hide password switch if needed"

wjt:
This reverts commit 2c7590a6f51c6c942c93a02fb8e26716efe42f36.

This was used only by Hack, and we are removing that configuration (https://github.com/endlessm/eos-image-builder/pull/899).

https://phabricator.endlessm.com/T29974


wjt:
Completely untested because the patch reverted cleanly.

PR #328, opened 2021-06-14
Move eos-test-mode to eos-boot-helper

GeorgesStavracas:
The script itself now lives in eos-boot-helper, which
installs it at a predefined location (/usr/sbin) and
thus we don't need to carry it anymore.

This must be squashed with a4201baf651 during the next
rebase.

https://phabricator.endlessm.com/T32045


dbnicholson (approved):
Looks good. I was going to wait until endlessm/eos-boot-helper#347 was ready, but I don't think there's any harm in merging this now.

PR #330, opened 2021-06-17
T31523 (Debian)

GeorgesStavracas:
No Debian packaging changes were needed.

https://phabricator.endlessm.com/T31523

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,3 +1,9 @@
+gnome-initial-setup (3.38.5-1endless1) eos; urgency=medium
+
+  * Endless 3.38.5 release
+
+ -- Georges Basile Stavracas Neto <georges@endlessos.org>  Thu, 17 Jun 2021 16:18:01 -0300
+
 gnome-initial-setup (3.38.0-1endless0) eos; urgency=medium
 
   * Endless 3.38.0 release


GeorgesStavracas:
retest this please

debian/changelog
```diff
@@ -1,3 +1,9 @@
+gnome-initial-setup (3.38.5-1endless0) eos; urgency=medium
+
+  * Endless 3.38.0 release
```
jprvita, line 3: I think you meant:
```suggestion
  * Endless 3.38.5 release
```
Also, as a side-note, we don't need to add any manual entries to debian/changelog for packages we track on git. Jenkins will always add a new entry on top based on the latest `Version_XYZ` / `Version_XYZ_debian` tags, and then count how many commits we have from the tags to the HEAD on the respective branches.

The only exception is when we have a downstream-only epoch (ex., pulseaudio), in which case we need to add one entry just to increase the epoch. But we should never have a downstream-only epoch (bad, bad pulseaudio).

jprvita (changes requested):

debian/changelog
GeorgesStavracas: Ouch, good catch. I always assumed bumping the Debian version in the changelog was considered a good practice, thanks for clarifying that it's not necessary though, I'll keep that in mind!

PR #331, opened 2021-07-23
Revert "debian: Change malcontent recommends to malcontent-data"

dbnicholson:
This reverts commit 3b0f7ff42e9df01897a44cb6acf70b1edefaa5f5. We're now
addressing the packaging delta with Debian and malcontent-data will no
longer exist.

https://phabricator.endlessm.com/T30746

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -11,14 +11,6 @@ gnome-initial-setup (3.38.0-1endless0) eos; urgency=medium
 
  -- Philip Withnall <pwithnall@endlessos.org>  Thu, 17 Sep 2020 17:00:00 +0100
 
-gnome-initial-setup (3.37.92-1endless1) eos; urgency=medium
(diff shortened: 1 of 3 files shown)


dbnicholson:
Marking this a draft as it shouldn't be merged until our updated malcontent has landed.

As for the content of this PR, this changes it back to the way debian has it, but that would add a dependency on the malcontent CLI, which isn't needed. If we adopt the changes I'm proposing in https://salsa.debian.org/freedesktop-team/malcontent/-/merge_requests/3, then this recommendation could be removed. On Endless we don't install Recommends, so the point is kind of moot, though.

pwithnall (approved):
LGTM.

dbnicholson:
Malcontent has rebuilt with the updated packaging. Merging.

PR #340, opened 2023-12-04
Update to GNOME Initial Setup 43.2-6 from Debian

wjt:
This does not carry forwards any of our previous changes. It is based on the upstream `gnome-43` branch and incorporates a number of patches from the Debian package, all of which were backports from newer GNOME versions and all of which are now on the upstream `gnome-43` branch. It then adds a few warning & crash fixes, which I will send upstream if approved here:

* 5ce147e9 goa: Simplify invoking helper synchronously
* 6ab4a51a timezone: Squash a possibly-uninitialised warning
* b3ada25e goa: Suppress a possibly-uninitialised warning
* e1ba44a4 driver: Avoid unused variable with webkit < 2.39.5

The idea is that this can supersede the 43.2-6 package that we temporarily imported to `eos:master:core`; and then we can reintroduce our bespoke changes incrementally.

This should be merged by force-pushing together with the packaging branch at #341:

```
git push --atomic origin +T35040-gnome-initial-setup-43:master T35040-gnome-initial-setup-43-debian:debian-master
```

### DIFF
--- a/gnome-initial-setup/pages/goa/gis-goa-page.c
+++ b/gnome-initial-setup/pages/goa/gis-goa-page.c
@@ -67,31 +67,20 @@ struct _ProviderWidget {
 };
 typedef struct _ProviderWidget ProviderWidget;
 
-G_GNUC_NULL_TERMINATED
 static char *
 run_goa_helper_sync (const char *command,
-                     ...)
+                     GError **error)
 {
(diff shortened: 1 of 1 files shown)


wjt:
retest this please

dbnicholson (approved):
LGTM. The fix patches look appropriate, too.

PR #341, opened 2023-12-04
Packaging for 43.2

wjt:
Packaging counterpart to #340.


dbnicholson (approved):
Seems fine to me. A couple notes for the future:

1. I personally wouldn't bother deleting the PKLA file. It's a few bytes in `/var` that will be never show up in the OS.
2. I've come to the conclusion that we shouldn't add `debian/changelog` entries. The information we put there is less useful than the commit log and it has a potential to cause conflicts with the `jenkins-obs` calculated version. As long as the `Version_*_debian` tag points to a commit containing Debian's entry in `debian/changelog` for the version we're modifying, then I think that's all that's needed. I did flatpak and gnome-control-center that way.

wjt:
Thanks, two great points. I think there is a high chance of a future version of this PR for Initial Setup 45 so I can address them there.

PR #344, opened 2023-12-21
Reintroduce most still-relevant downstream changes atop 43

wjt:
This reintroduces the majority of our downstream functionality from eos5.1 that we want to preserve. There are 3½ conspicuous gaps:

- factory dialog, accessed by pressing Ctrl+f on the first page
- live boot support & invoking eos-installer on eosinstaller images (which are entwined)
- the eula page
- transifex goop

I'll bring these back in later PRs but they are more involved than the other changes, due to the GTK 4 port.

I first rebased our existing changes onto the latest 41.x point release, not because I plan to merge that rebase but because helped apply them to this new baseline. I have made notes on what I dropped, squashed, and split on the following PR:

- https://github.com/endlessm/gnome-initial-setup/pull/343

So this PR is the result of rebasing that series onto `master`. It depends on the following:

- [ ] https://github.com/endlessm/gnome-initial-setup/pull/342

Picked
------

5d81bdcc copypasta code: Use GTask in UmRealmManager

- No conflicts.

a2672e30 Allow initial setup to switch metrics on and off

- Applied without conflicts

539bb54f language: List 'endless' in the id-to-logo mapping array

- Applied without conflicts

18d4a1e3 account-local: limit full name to 80 characters

- Applied without conflicts

23f1bbcd account-local: username suggestions

- Resolved one minor conflict due to GTK 4 port.

bfd80ac4 page-util: Add API to read the EOS image version

- Applied without conflict

d38138f8 driver: Track details of current image & boot type

- Fixed minor conflicts

930c0fea input-chooser: sort default layout before others

- Applied without conflicts.

954c2c21 input-chooser: consider only language or country layouts non-extra

- Fixed minor conflicts.
- It's actually pretty hard to find a locale with fewer than 5 language/country
  layouts. Macedonian is one.
- This patch actually has no effect – choose_non_extras() is already called
  before adding the massive list of other layouts - but it is cleaner. So should go upstream. Rewording commit message accordingly.

1c44d429 input-chooser: Always show the selected layout

- Applied without conflicts
- I'm not 100% sure this patch does anything though

384f5c91 keyboard: add us/us+intl default layouts as fallback

- Applied without conflicts

8cc19caa language-chooser: make it more clear for the users

The previous version of this patch professes to do the following:

1. Vertically centre the initially-selected language – not needed, the initial
   language is always in the group that sits “above the fold”. I think an
   earlier iteration just showed all languages in alphabetical order rather than
   showing the cherry-picked list of 8 languages plus the current locale.

2. Show the searchbar by default – no longer needed, the search bar is always
   shown by default.

3. Show all languages by default because it is not clear that the
   three-vertical-dots row means "show the other languages". I have to say I am
   sceptical of this because the original report that led to this patch reads:
   https://phabricator.endlessm.com/T15310 
   
   > We only show "English (United States)" in the first screen of the FBE on a
   > "base" image, which is our most commonly downloaded product.”
   
   That's not true any more: we show 8 locales, plus the currently-selected
   locale if it is not one of those.
   
   But anyway. It is easy enough to carry forward this specific behaviour
   change - just add a single function call.

pick d39934f8 timezone: set default 12h/24h from locale

Keep for now. Upstream this has been resolved in https://gitlab.gnome.org/GNOME/gsettings-desktop-schemas/-/merge_requests/55 for GNOME 46 but for now let's just keep our workaround.

pick a1051474 language: allow to specify initial languages from vendor.conf

Applied cleanly. Amended the patch to remove two unused variables, left over
from a simplification in a previous rebase. This would be interesting to send
upstream, though since there is a corresponding setting in gnome-control-center
it would actually be better as a GSetting.

pick f88cfa8e language: make sure initial language is always at the top

Applied cleanly. Doesn't seem *essential* but is a nice improvement.

pick bbee6361 driver: Inhibit the idle state

Applied cleanly. Keeping this for now but if we update to Initial Setup 45 a
better fix is present there.

pick dee629ff account-local: add toggle for passwordless accounts

Fixed conflicts from GTK 4 port & updated for GTK 4 API/.ui file syntax changes.

pick 6fc61aa7 account,password: Hide password page for passwordless account

Fixed conflicts from GTK 4 port & updated for GTK 4 API changes.

pick 76eef865 keyboard: manually pass locale to gkbd-keyboard-display

Applied without conflicts. This patch is weird: it builds a command-line as
a single string then parses it with `g_shell_parse_argv`. Why not just build an
array of strings? I think I have a patch to do this better lying around somewhere
so I will upstream that instead later.

pick 51921a3d timezone: Hide when automatically detected

Applied without conflict.

pick b625e8d0 eos: Report metrics from parental controls page

Applied cleanly. Amended to introduce the eosmetrics-0 dependency (which would
otherwise be added later in the series by the EULA page)

Dropped
-------

drop 99274c42 copypasta code: Use icon name instead of stock icon

- Already effectively present upstream as part of GTK 4 port

drop 65a604bb Add product_serial.conf file

- We will stop displaying a Gigabyte-specific subset of the product UUID (not
  serial despite what the commit message claims)
  
drop 77bb2e93 password: Show the last character when entering the password

- This is now set in eos-theme

drop 475b5102 input-chooser: allow CcInputChooser to expand

- No longer relevant after GTK 4 port

drop a6334008 Change generated avatar for empty name

- Removed this change – I think the upstream default looks better

TODO
----

7e692863 Allow pages to specify a GtkAccelGroup
7f409293 factory test mode: implement the factory dialog

e2f1e865 page-util: Add logic to launch eos-installer
1a94089c language: Launch eos-installer on eosinstaller images

5c024d34 live-chooser: add the Live Chooser page
52826c5c pages: Don't show some pages on live sessions
9d8c30d8 summary: show different messages on live sessions

1617c1e2 endless-eula: Add endless eula page

0aca1ee0 Configuration for Transifex
a391e024 (HEAD -> T35040-rebase-onto-41.4, origin/T35040-rebase-onto-41.4) Add translation support for several new languages

### DIFF
--- a/data/20-gnome-initial-setup.rules.in
+++ b/data/20-gnome-initial-setup.rules.in
@@ -10,6 +10,7 @@ polkit.addRule(function(action, subject) {
         return undefined;
 
     var actionMatches = (action.id.indexOf('org.freedesktop.hostname1.') === 0 ||
+                         action.id === 'com.endlessm.Metrics.SetEnabled' ||
                          action.id.indexOf('org.freedesktop.NetworkManager.') === 0 ||
                          action.id.indexOf('org.freedesktop.locale1.') === 0 ||
                          action.id.indexOf('org.freedesktop.accounts.') === 0 ||
--- a/gnome-initial-setup/cc-common-language.c
+++ b/gnome-initial-setup/cc-common-language.c
@@ -34,6 +34,8 @@
 
 #include "cc-common-language.h"
 
+#include "gnome-initial-setup.h"
+
 static char *get_lang_for_user_object_path (const char *path);
 
 static char *current_language;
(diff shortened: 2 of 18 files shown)


dylanmccall (approved):
I looked through these patches and all seems well / builds and runs correctly for me :+1:

PR #345, opened 2024-02-15
Reimplement factory dialog

wjt:
This is spiritually a forward-port of 7e692863 and 7f409293, but is in practice almost a total rewrite:

- Port to GTK 4.
- Use GtkShortcut for keyboard shortcut, in place of removed
  GtkAccelGroup API. We cannot use gtk_widget_class_add_binding on the
  GisLanguagePageClass because we apparently need to handle the event in
  the CAPTURE phase not the BUBBLE phase if we want it to trigger when
  the Next button in the headerbar (which is outside the
  GisLanguagePage) has focus. I learned this by reading the
  implementation of mnemonics in GtkLabel and copying it here.
- Rewrite the dialog as a custom widget, which trades one kind of
  boilerplate for another.
- Use GtkWindow not GtkDialog which is deprecated in GTK 4.10. Debian
  Bookworm has 4.8 but we were not using any of the special
  functionality of GtkDialog anyway.
- Don't show ARM product ID or x86_64 (vendor-specific selection of
  bytes from) product_uuid – no longer needed.
- Use PRETTY_NAME from os-release rather than manually combining NAME
  and VERSION to create the same string.

I also cherry-picked an upstream patch to fix a crash which you hit when running the app with `$LANG` unset, and included a couple of commits that fixup previous Endless-specific changes because I couldn't resist making some simplifications as I wrote this.

Screenshot from 2024-02-15 13-37-22

### DIFF
--- a/data/20-gnome-initial-setup.rules.in
+++ b/data/20-gnome-initial-setup.rules.in
@@ -11,6 +11,7 @@ polkit.addRule(function(action, subject) {
 
     var actionMatches = (action.id.indexOf('org.freedesktop.hostname1.') === 0 ||
                          action.id === 'com.endlessm.Metrics.SetEnabled' ||
+                         action.id === 'com.endlessm.TestMode' ||
                          action.id.indexOf('org.freedesktop.NetworkManager.') === 0 ||
                          action.id.indexOf('org.freedesktop.locale1.') === 0 ||
                          action.id.indexOf('org.freedesktop.accounts.') === 0 ||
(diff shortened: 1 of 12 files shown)


dylanmccall (approved):
This builds and runs for me, and it all looks good to me :) I noticed the `serial-label` label has been removed here. Is that what you're referring to with "don't show ARM product ID or x86_64 product_uuid"? (I didn't know about that label until today, so I don't know the relevant terminology).

wjt:
Yes, despite its name the "serial" was not actually the serial number. It was a selection of bytes, formatted as hexadecimal, from the product_uuid, which was added 7+ years ago for the OEM producing (I believe) the Mission One.

PR #346, opened 2024-02-15
Revert "Build-depend on libattr1-dev"

wjt:
This reverts commit 042b5dd61c633a000933b5d269397ed119adf12f.

6288f0e6f39b5c1c391edb6701bbf9966bc6b9a6 on the code branch reimplements the relevant function using GFile.

Depends on:

- [x] #345


dylanmccall (approved):
Looks like a good change to me, and makes sense to do :) We should probably remember to change the commit message once the relevant commits are rebased.

wjt:
Which commit message would we change? I imagine we would drop the commit + its revert

PR #347, opened 2024-02-16
privacy: Add metrics toggle

wjt:
This is the spiritual successor of half of
1617c1e20c16c2dca7abd2401cf357c7d54ff389 (“endless-eula: Add endless eula page”). I believe the setting fits more naturally on the Privacy page, where it can be displayed more prominently; this will also allow more vertical space to read the EULA in the EULA page.

The manual definition of an autocleanup for GisPrivacyPage will be obsolete in GNOME 46, where I have rewritten this file to use G_DEFINE_FINAL_TYPE which includes that boilerplate.

I have kept the strings from the previous implementation. “Help make Endless better for everyone” looks a bit weird next to the location services option on this page, but I would rather keep the wording as-is and then haggle over it separately. (For upstream, it would need to be adjusted anyway to interpolate the OS vendor at runtime.)

Compared to the previous implementation, I made the proxy construction async, because it's barely harder to do that than to make it synchronously. I kept the synchronous call to SetEnabled() because that's what everything else does in the save_data hook.

There is an argument that we should instead call this from gis_privacy_page_apply(), similar to the other toggles. Either one would work, and actually I wrote that first. gis_privacy_page_apply() can also be async. But if you make it async, a spinner and cancel button are shown; but because SetEnabled() is really quick, they only flash into existence for a frame before being replaced by the next page. It looks bad.

I can't remember why we previously deferred actually changing the option to save_data, but let's just leave it there for now.

## Normal screenshot

image

The text is obnoxiously small, right? But that is not new in this commit.

## Large text screenshot

image

Ahh, so much more legible!

### DIFF
--- a/gnome-initial-setup/pages/privacy/gis-privacy-page.c
+++ b/gnome-initial-setup/pages/privacy/gis-privacy-page.c
@@ -38,15 +38,21 @@
 
 #include "gis-page-header.h"
 
+G_DEFINE_AUTOPTR_CLEANUP_FUNC(GisPrivacyPage, g_object_unref)
+
 struct _GisPrivacyPagePrivate
 {
   GtkWidget *location_switch;
   GtkWidget *reporting_group;
   GtkWidget *reporting_label;
   GtkWidget *reporting_switch;
+  GtkWidget *metrics_group;
+  GtkWidget *metrics_switch;
   GSettings *location_settings;
   GSettings *privacy_settings;
   guint abrt_watch_id;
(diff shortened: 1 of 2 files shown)


dylanmccall (approved):
Works for me, and seems just right! Looking at this, I'm kind of wishing our Metrics panel in gnome-control-center aligned with this better (same style switch, same language, etc.). But hopefully that will be moot when we're switching to an upstream Metrics system in the future :b

PR #348, opened 2024-02-16
Add eos-installer patches for GNOME 43

dylanmccall:

-----

+ e2f1e865 page-util: Add logic to launch eos-installer
Fairly straightforward. Removed some unnecessary imports and stopped using gtk_dialog_run as in other places, and changed `gis_page_util_run_reformatter` to always return after running into an error.

+ 1a94089c language: Launch eos-installer on eosinstaller images
Also pretty simple.

I haven't tested this very thoroughly at all, but it seems to be working as expected.


wjt (approved):
LGTM!

PR #349, opened 2024-02-23
GNOME 43 live USB patches

dylanmccall:


dylanmccall:
This is my update of the live-chooser page, taking what we had before with some small tweaks, replacing the (no longer available) GtkFrame with a separator:

Screenshot from 2024-02-22 21-17-44

There's a weird thing going on with GtkPicture and these layout containers where `install-endless.png` ends up getting way more vertical space allocated to it than it needs, resulting in a huge gap. I worked around that for the moment by, strangely, adding a bunch of empty space to that image. There still is some weird unexplained extra spacing, and if you can figure out _why_ I'll be very grateful, but otherwise it's probably not too big a deal.

(I have a weird feeling here that making an SVG for the same image, and using GtkImage and `<property name="pixel_size">96</property>` as in the welcome page, would probably be the optimal solution, but I don't think it's necessary to deal with that right now).

dylanmccall:
Here's our updated Summary page:

Screenshot from 2024-02-23 13-47-18

(Don't mind the mismatched "Start Using Fedora Linux" label :b).

I tweaked the warning label to not have bold text because I felt it was too shouty with the new bold title text from `AdwStatusPage`. Unfortunately this was originally done using Pango markup in the label itself, so it's a string change, but I think we have a couple other string changes anyway and this one's a pretty easy find & replace.

For reference, here's the shouty verison:

Screenshot from 2024-02-23 13-47-32

starnight (approved):
Thanks

dylanmccall:
> I notice that the live chooser page has a (greyed-out?) Next button. The old version of these patches added a `GisPage:has-forward` boolean that `GisAssistant` would use to decide whether to hide the Next button.

Oh yeah, and we set that property but I see how binding it to the button's visibility would have gotten lost. I should have a PR fixing that in a few minutes.

dylanmccall:
Okay @wjt, I did that over here: https://github.com/endlessm/gnome-initial-setup/pull/352.

It looks like the page has had this behaviour for at least the previous Endless OS release. The has-forward property is a little bit counter-intuitive :)

PR #350, opened 2024-02-23
Cleanups for timezone page

wjt:
Some polish for the timezone page, plus a fix for a downstream change.

The geoclue change is [already upstream](https://gitlab.gnome.org/GNOME/gnome-initial-setup/-/commit/06544c44fd12d57a1b05c13987cd9abe178eb3cf). The `G_DECLARE_FINAL_TYPE` and checked-cast changes are sitting on [this MR](https://gitlab.gnome.org/GNOME/gnome-initial-setup/-/merge_requests/242) which I will merge if they get a positive review here. They applied with some small conflicts due to downstream changes to this page.


dylanmccall (approved):
I always remember useful things I should be doing when I review these cleanups :) I read through this and spotted nothing amiss.

PR #352, opened 2024-02-26
fixup! live-chooser: add the Live Chooser page

dylanmccall:
Hide the "Next" button in the Live Chooser page. It appears that the original intent in gis-live-chooser-page.c was for this this button to be hidden. However, `gis_live_chooser_page_constructed` was incorrectly setting the `has-forward` property to FALSE (its default value), instead of setting it to TRUE.


wjt (approved):

wjt:
Wow, very confusing.

PR #353, opened 2024-02-27
Readd Endless EULA page

wjt:
This is a single commit stacked on top of:

- #351

## Spanish

Spanish terms of use embedded in Initial Setup

(Ignore the lack of translations on the headers, and feast your eyes on the Spanish-language PDF.)

## English

Screenshot from 2024-02-27 21-18-36

## English large text

Screenshot from 2024-02-27 21-18-43

Something is funky about the font sizes in the PDF.js toolbar. It's not very visible here but it's really obvious if you open the dropdown menu. However… it's legible, and not a problem for today.

## License list overlay (large text)

image

## EOS 5.1, for reference

Screenshot from 2023-12-11 16-38-58

I think that allowing the PDF to extend to the full width of the window, and giving it more vertical height by moving the metrics toggle to the privacy page, makes a big difference in terms of legibility. I personally find it almost impossible to read the EOS 5.1 terms.

## Changes in this version of the patch

Compared to its predecessor 1617c1e20c16c2dca7abd2401cf357c7d54ff389:

- Eliminate use of Evince library to display PDFs, as it has no GTK 4
  version at present. Use WebKitGTK's built-in copy of PDF.js instead.
  This means we gain a somewhat ugly toolbar, but there's no way to
  disable it short of embedding our own PDF.js-based viewer, which is
  not worth doing. And actually it's kind of nice to get search and zoom
  for free!

- Remove metrics control – this now lives on the privacy page, which
  seems a more natural home.

- Remove reloading the EulaView when the locale changes. (The
  locale_changed hook is actually mostly useless because all pages after
  the currently-selected page are recreated from scratch when the locale
  changes.)

- As a result, GisEndlessEulaPage is now almost nothing but boilerplate
  and a template.

- Preserve the inability to browse to any page other than file:/// URIs
  (which in practice means the English-language terms of use, from other
  translations) and the license server on localhost. You might think
  this is not a reasonable restriction because there are links to other
  pages from the terms. But as explained in the comment it may actually
  be desirable because the user is offline.

### DIFF
--- a/gnome-initial-setup/pages/endless-eula/gis-endless-eula-viewer.c
+++ b/gnome-initial-setup/pages/endless-eula/gis-endless-eula-viewer.c
@@ -1,0 +1,298 @@
+/* -*- mode: C; c-file-style: "gnu"; indent-tabs-mode: nil; -*- */
+/*
+ * Copyright (C) 2018–2024 Endless OS Foundation LLC
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License as
+ * published by the Free Software Foundation; either version 2 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful, but
+ * WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA
+ * 02111-1307, USA.
+ *
+ * Written by:
+ *     Damián Nohales <damian@endlessm.com>
+ *     Will Thompson <wjt@endlessos.org>
+ */
+
+#include "config.h"
+#include "gis-endless-eula-viewer.h"
+
+#include <gtk/gtk.h>
+
+struct _GisEndlessEulaViewer {
+  WebKitWebView parent;
+
+  char *terms_uri;
+};
+
+G_DEFINE_TYPE (GisEndlessEulaViewer, gis_endless_eula_viewer, WEBKIT_TYPE_WEB_VIEW);
+
+#define LICENSE_SERVICE_URI "http://localhost:3010"
+
+static void
+notify_progress_cb (GObject    *object,
+                    GParamSpec *pspec,
+                    gpointer    user_data)
+{
+  GtkWidget *progress_bar = GTK_WIDGET (user_data);
+  WebKitWebView *web_view = WEBKIT_WEB_VIEW (object);
+  gdouble progress;
+
+  progress = webkit_web_view_get_estimated_load_progress (web_view);
+
+  if (progress == 1.0)
+    gtk_widget_hide (progress_bar);
+  else
+    gtk_widget_show (progress_bar);
+
+  gtk_progress_bar_set_fraction (GTK_PROGRESS_BAR (progress_bar), progress);
+}
+
+static void
+notify_title_cb (GObject    *object,
+                 GParamSpec *pspec,
+                 gpointer    user_data)
+{
+  WebKitWebView *web_view = WEBKIT_WEB_VIEW (object);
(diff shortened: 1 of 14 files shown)


gnome-initial-setup/pages/endless-eula/gis-endless-eula-viewer.c
```diff
@@ -0,0 +1,298 @@
...
+{
+  g_autoptr(GFile) file = NULL;
+
+  file = get_terms_document ();
+  if (file == NULL)
+    return;
+
+  g_assert (self->terms_uri == NULL);
+  self->terms_uri = g_file_get_uri (file);
+  webkit_web_view_load_uri (WEBKIT_WEB_VIEW (self), self->terms_uri);
```
dylanmccall, line 255: Right now we aren't showing anything if the URI fails to load, but this is not a regression from the previous version. However, this document does take a noticeable time to load. Should we add a GtkOverlay with a progress bar here, similar to the one in `show_uri_in_modal`?

dylanmccall (commented):
It's a shame about those Print and Save buttons in the PDF viewer: they aren't hooked up to anything, so they're just there, ready to confuse people. (I really dislike that Webkit does this kind of thing where it's stuffed full of features and default behaviours designed around full fat web browsers while pretending to be a simple unopinionated library for displaying web content). Probably not worth the bother of implementing the appropriate hooks. I think this is a pretty good approach, and I agree with the links not opening in a browser since they're written out in full in the document anyway.

There are a couple spots I'm curious about, but I read through this and tested it out and it seems to be in pretty good shape.

gnome-initial-setup/pages/endless-eula/gis-endless-eula-viewer.c
```diff
@@ -0,0 +1,298 @@
...
+      webkit_policy_decision_use (decision);
+      return TRUE;
+    }
+
+  webkit_policy_decision_ignore (decision);
+
+  const gchar *title = NULL;
+
+  if (g_str_has_prefix (uri, "file:"))
+    title = _("Terms of Use");
```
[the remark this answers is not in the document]
wjt: I don't really think this matters. We only show Initial Setup in the locked-down kiosk so you can't get to a file to drag in. And there is no security boundary here – just trying to avoid having to deal with unintentional navigation away from the terms of use.

gnome-initial-setup/pages/endless-eula/gis-endless-eula-viewer.c
wjt: We could. This page is loaded in the background, while the user is looking at the language & keyboard layout (and possibly live-chooser) pages, but I guess we can do that anyway.

manuq:
As a visual impaired person, I welcome the zoom option! Accesibility beats ugliness.

gnome-initial-setup/pages/endless-eula/gis-endless-eula-viewer.c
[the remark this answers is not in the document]
wjt: Also this is the popover where no attempt is made to prevent further navigation. I think it's fine because you can just close the window if you get stuck .

wjt:
I think the main shortcoming is the clutter from the PDF.js toolbar, and the way to address that would be to convert the terms of use to HTML, and ensure that the text is the same size as the GTK body text size.

I started experimenting with this at https://github.com/endlessm/eos-license-service/pull/28 and I think it is viable. But I would prefer to do that as a separate step.

Another option could be to inject some CSS into the webview to hide the PDF.js toolbar, but I am inclined not to do that because the PDF does not follow the system font size so zoom is a useful a11y feature (as mentioned above).

So personally I would be inclined to land this as-is, and if we feel the need to further improve the presentation go the HTML route. WDYT?

dylanmccall:
That sounds like a good plan to me. Having the terms of use as a plain HTML file would solve the slowish loading issue, too. I'm happy with merging this as-is and hopefully improving it later around that proposal.

gnome-initial-setup/pages/endless-eula/gis-endless-eula-viewer.c
wjt: I tested this further and you're right, the PDF is actually only loaded once the page is displayed. I think it's fast enough but a progress bar might be nice as a future iteration.

PR #359, opened 2024-12-10
privacy: Add BeaconDB privacy policy link

wjt:
Requires a corresponding GeoClue config change to use BeaconDB.

image

https://phabricator.endlessm.com/T35586


wjt:
I added this as a separate label so that it doesn't destroy the translations for the existing sentence.

starnight (approved):
LGTM!

